PR #35, opened 2017-12-07
Create integration tests

lnyng:


tylerharter:
Leon confirmed we don't need this anymore.

PR #45, opened 2018-08-04
Add i.mutex.Lock() before delete(i.pkgStates, pkg) to avoid calling i…

gtotoy:
Add i.mutex.Lock() before delete(i.pkgStates, pkg) to avoid calling i.mutex.Unlock() twice when install package cmd returns an error.


edoakes:
Looks good. One unrelated issue that it'd be great if you could fix too: the entry should be RLocked in both the creation and reading cases. Could you move the `RLock()` call outside of the `else` statement?

PR #53, opened 2019-04-22
Work of Lambda Team in S19 CS838 Edge Computing Lab

gtbai:
**This PR has been split into the following two PRs: #54 (code only) and #55 (vendor only).**

In this PR, we added the following two features for workers:

1. Enable requests of the same lambda to reuse a handler.
    + To accomplish this, we moved the code that adds a new handler to the `HandlerManagerSet .HandlerManager.handlers` list.

2. Added request queuing mechanism (mem usage hard limit) for workers.
    + Added a utility function to get the current machine memory usage, making use of [shirou/gopsutil](https://github.com/shirou/gopsutil)
    + Added both a soft limit (`Config.Create_sandbox_soft_limit `/`HandlerManagerSet.sbSoftLimit`) and a hard limit (`Config.Create_sandbox_hard_limit `/`HandlerManagerSet.sbHardLimit`) for the memory usage when creating new sandboxes.
        + When memory usage exceeds the soft limit, at most `Config.Max_num_sandbox_creating `/`HandlerManagerSet.maxNumSbCreating ` sandboxes can be being created at the same time.
       + When memory usage exceeds the hard limit, requests creating new sandboxes will be blocked, until memory usage decreases.


edoakes:
Thanks for the contributions! A few questions/suggestions:

1. What exactly is the problem you're fixing related to handler reuse? From what I remember, requests to the same lambda should already be reusing the same handler - was this not working in some circumstances?
2. What is the configurable memory usage limit a percentage of? The total memory available on the host? I think it would be better to have a concrete limit in bytes instead, as that's standard in most cluster schedulers (e.g., [kubernetes](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory)).
3. Some tests for both of these would be great :)

gtbai:
> Thanks for the contributions! A few questions/suggestions:
> 
> 1. What exactly is the problem you're fixing related to handler reuse? From what I remember, requests to the same lambda should already be reusing the same handler - was this not working in some circumstances?
> 2. What is the configurable memory usage limit a percentage of? The total memory available on the host? I think it would be better to have a concrete limit in bytes instead, as that's standard in most cluster schedulers (e.g., [kubernetes](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory)).
> 3. Some tests for both of these would be great :)

@edoakes Thanks for the comments!

1.  We made this change because OpenLambda crashed when we experimented with it on small VMs with only 1GB memory on AWS. The reason is that when multiple requests of the same lambda arrive together, the [current code](https://github.com/open-lambda/open-lambda/blob/6df65ce3e967baac8a8770286caa27ae0fcf5963/worker/handler/handler.go#L150) will create a handler for each of them, with a sandbox created for each later. This will drain out the limited memory and lead OpenLambda to a crash. 

    As we discussed in class, we could have heterogeneous machines in the edge cluster, some of which might have a small amount of memory. Therefore, we adjusted the workflow to push a handler into the queue immediately after it has been created, which enables requests of the same lambda to reuse the handler as well as the sandbox. With this, memory is saved and OpenLambda does not crash anymore. 

    When we experimented on [CloudLab c220g2 nodes](http://docs.cloudlab.us/hardware.html) with 160GB memory, the issue does not appear.

2. Yes, it is a percentage of the **total memory of the host**. We chose to use a relative limit here because we were talking a lot about the potential heterogeneity of machines in the edge cluster in our class, so we supposed a relative limit for the memory would make OpenLambda easier to config and use out of the box.  

    That said, we did check the kubernetes link you attached, and agreed that it may be better to follow the standard way. 

    @tylerharter Could you share your opinion on this?

3. Could you share with us an entry where we could add more unit test? 😁

dennisszhou:
>    When we experimented on [CloudLab c220g2 nodes](http://docs.cloudlab.us/hardware.html) with 160GB memory, the issue does not appear.
> 2. Yes, it is a percentage of the **total memory of the host**. We chose to use a relative limit here because we were talking a lot about the potential heterogeneity of machines in the edge cluster in our class, so we supposed a relative limit for the memory would make OpenLambda easier to config and use out of the box.

FWIW, without reading the code, I agree with Ed on the hard limit. Percentages really mean nothing without context of a host. If I have a config for 1% of a host, that may not boot a jvm on the AWS instance, while it may boot on the cloudlab machine.

Again I'm sorry I don't have time to read the code, but I imagine queueing should be based off of total consumption of a host (sum up the lambda usage) and then you have the discussion of how to be work conserving (oversubscription of memory and how to deal with memory pressure if everyone wants all the memory they ask for).

edoakes:
I see now, thanks for the clarification. Essentially your change is a performance tradeoff of memory for latency (as you will now queue some of the incoming requests instead of executing them in parallel). As neither of these options is clearly better in every situation, it would be better to have this as a configurable option. Since we don't currently have any intelligent scheduling, I would suggest an integer limit for the number of concurrent handlers per lambda - spin them up concurrently to that limit, then queue any remaining requests.

worker/config/config.go
```diff
@@ -127,6 +134,20 @@ func (c *Config) Defaults() error {
 		return fmt.Errorf("must specify local registry directory")
 	}
 
+	// mem usage limits for creating sandboxes
+	if c.Create_sandbox_soft_limit == 0.0 {
+		c.Create_sandbox_soft_limit = 0.5
```
tylerharter: Can't we push the default soft limit closer than this?  Say 0.8?

worker/go.mod
```diff
@@ -1,16 +1,13 @@
...
 require (
-	github.com/fsouza/go-dockerclient v1.3.3
-	github.com/go-ini/ini v1.41.0 // indirect
-	github.com/gopherjs/gopherjs v0.0.0-20181103185306-d547d1d9531e // indirect
-	github.com/jtolds/gls v4.2.1+incompatible // indirect
-	github.com/minio/minio-go v6.0.12+incompatible
-	github.com/mitchellh/go-homedir v1.0.0 // indirect
-	github.com/smartystreets/assertions v0.0.0-20180927180507-b2de0cb4f26d // indirect
-	github.com/smartystreets/goconvey v0.0.0-20181108003508-044398e4856c // indirect
+	github.com/StackExchange/wmi v0.0.0-20181212234831-e0a55b97c705 // indirect
```
tylerharter: why are you changing this?  Any re-sync of the vendors should be its own PR.

tylerharter: Otherwise a 289-file change is just too much.

worker/handler/handler.go
```diff
@@ -117,13 +124,32 @@ func NewHandlerManagerSet(opts *config.Config) (hms *HandlerManagerSet, err erro
...
 	hms.lru = NewHandlerLRU(hms, opts.Handler_cache_size) //kb
 
 	return hms, nil
 }
 
+var memUsage float64 = 0.0
+var memLastUpdateTime time.Time = time.Now()
+
+func getMemUsage() float64 {
+	if (time.Since(memLastUpdateTime)) > time.Second {
```
tylerharter: Either make this configurable, or add a TODO so we don't forget that we hard-coded the freq to time.Second

worker/handler/handler.go
```diff
@@ -277,6 +309,16 @@ func (h *Handler) RunStart() (ch *sb.Channel, err error) {
 
 	// create sandbox if needed
 	if h.sandbox == nil {
+		// if memUsage exceeds the sbSoftLimit, we would create at most maxNumSbCreating
+		// sandboxes at the same time; if memUsage exceeds the sbHardLimit, we would block 
+		// all the creation requests until there are more free mem.
+		hms.sbCreatingCond.L.Lock()
+		for (getMemUsage() > hms.sbSoftLimit && hms.numSbCreating > hms.maxNumSbCreating) || getMemUsage() > hms.sbHardLimit {
+			hms.sbCreatingCond.Wait()
```
tylerharter: Shouldn't there be some kind of timeout here?  You only broadcast to this cond var in two places:
1. after a container has started
2. after a container has been "nuked"

Even if the set of containers isn't changing, it's possible the existing containers could release some of the memory they've been using, in which case we'd want to start more containers.  Since there's no way for us to be notified when a container starts using less memory, I think on some level we need to rely on polling.

Maybe you can add a timeout for your cond variable?

worker/config/config.go
```diff
@@ -127,6 +134,20 @@ func (c *Config) Defaults() error {
...
+	} 
+
+	if c.Create_sandbox_hard_limit == 0.0 {
+		c.Create_sandbox_hard_limit = 0.9
+	} 
+
+	// max number of sandboxes being created at the same time (after hit soft limit)
+	if c.Max_num_sandbox_creating == 0 {
+		c.Max_num_sandbox_creating = 4
```
tylerharter: What's the rational for this limit?

worker/go.mod
gtbai: This should not have been changed, I just happened to re-vendor all the packages... sorry about the confusion.

I have split this PR into two PRs: https://github.com/open-lambda/open-lambda/pull/54 (code only) and https://github.com/open-lambda/open-lambda/pull/55 (vendor only), and corrected the vendor sync issue in the second one.

worker/config/config.go
gtbai: We can certainly do this.

worker/handler/handler.go
gtbai: Would update.

worker/handler/handler.go
gtbai: A timeout makes sense, would update.

worker/config/config.go
gtbai: We are thinking to reduce the concurrency of spinning up new sandboxes once the memory usage hits `Create_sandbox_soft_limit`.

Suppose without `Create_sandbox_soft_limit ` and `Max_num_sandbox_creating `, when the memory usage is close to the `Create_sandbox_hard_limit `, a bunch of incoming requests would detect that the memory usage is lower than the hard limit, thus try to spin up multiple sandboxes concurrently, leading to a surge in memory usage and crash the system.

In addition, this limit is configurable in twofold: It is configurable in itself, plus that once the soft limit is set to be greater than or equal to the hard limit, it does not take effect anymore.

worker/handler/handler.go
tylerharter: Great, looking forward to seeing this added in the other PR.

worker/config/config.go
tylerharter: Let me paraphrase what I think you're saying: mem measurements are only updated every second, so we don't want to risk letting too many in at once, shooting past the limit before we re-measure.

If that's correct, I think the correct thing to do is limit the rate, not the concurrency.

How about we do this:
1. estimate the memory consumption of the new sandbox (based on a config, or average of all other lambdas)
2. update the mem-usage snapshot based on our estimate (it will be exact in the next second)
3. wait if starting would bump our mem estimate past the threshold

worker/handler/handler.go
gtbai: Updated in https://github.com/open-lambda/open-lambda/pull/54.

worker/config/config.go
gtbai: Your understanding is correct.

Your proposal makes a lot of sense, though we think it may cause in some cases, for example: suppose a sandbox creating request comes, with the mem usage is right under hard limit, we let it spin up a sandbox and "update the mem-usage snapshot based on our estimate" over the hard limit, then after one sec another sandbox creating request comes, it will remeasure the exact mem usage -- if the first sandbox is still being created and takes little memory, then this request will be approved as well, leading mem usage to past the hard limit.

To resolve this, we are thinking to keep the [`hms.numSbCreating`](https://github.com/open-lambda/open-lambda/pull/54/files#diff-3d8711386f9437d690ff9ee2452d22b2R353) counter to keep track of how many sandboxes are being created, and approve/queue a request based on [whether the sum of host mem usage and estimated mem usage of all sandboxes being created is over the hard limit](https://github.com/open-lambda/open-lambda/pull/54/files#diff-3d8711386f9437d690ff9ee2452d22b2R344).

We have updated this in https://github.com/open-lambda/open-lambda/pull/54.

gtbai:
> 2. What is the configurable memory usage limit a percentage of? The total memory available on the host? I think it would be better to have a concrete limit in bytes instead, as that's standard in most cluster schedulers (e.g., [kubernetes](https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#meaning-of-memory)).

> FWIW, without reading the code, I agree with Ed on the hard limit. Percentages really mean nothing without context of a host. If I have a config for 1% of a host, that may not boot a jvm on the AWS instance, while it may boot on the cloudlab machine.

We have [updated the hard limit to a concrete limit in kb](https://github.com/open-lambda/open-lambda/pull/54/files#diff-ee991a7139f03fd035fe8ef4a4a1b151R53) (to be consistent with `Config.Handler_cache_size` and `Config. Import_cache_size `), thanks @edoakes @dennisszhou for the suggestions!

tylerharter:
I've merged in the separate vender PR.  Closing this because all the work is being done in #54 now.

PR #54, opened 2019-04-25
[Code Only] Work of Lambda Team in S19 CS838 Edge Computing Lab

gtbai:
This is the code part of PR https://github.com/open-lambda/open-lambda/pull/53.

> In this PR, we added the following two features for workers:
> 
> 1. Enable requests of the same lambda to reuse a handler.
>    
>    * To accomplish this, we moved the code that adds a new handler to the `HandlerManagerSet .HandlerManager.handlers` list.
> 2. Added request queuing mechanism (mem usage hard limit) for workers.
>    
>    * Added a utility function to get the current machine memory usage, making use of [shirou/gopsutil](https://github.com/shirou/gopsutil)
>    * Added both a soft limit (`Config.Create_sandbox_soft_limit `/`HandlerManagerSet.sbSoftLimit`) and a hard limit (`Config.Create_sandbox_hard_limit `/`HandlerManagerSet.sbHardLimit`) for memory usage when creating new sandboxes.
>      
>      * When memory usage exceeds the soft limit, at most `Config.Max_num_sandbox_creating `/`HandlerManagerSet.maxNumSbCreating ` sandboxes can be being created at the same time.
>      * When memory usage exceeds the hard limit, requests creating new sandboxes will be blocked, until memory usage decreases.

### DIFF
--- a/worker/config/config.go
+++ b/worker/config/config.go
@@ -46,6 +46,13 @@ type Config struct {
 	Handler_cache_size int `json:"handler_cache_size"` //kb
 	Import_cache_size  int `json:"import_cache_size"`  //kb
 
+	// mem usage limits for creating sandboxes
+	Create_sandbox_soft_limit float64 `json:"create_sandbox_soft_limit"` //mem usage percentage, from 0.0 (0%) to 1.0 (100%)
+	Create_sandbox_hard_limit float64 `json:"create_sandbox_hard_limit"` //mem usage percentage, from 0.0 (0%) to 1.0 (100%)
+
+	// max number of sandboxes being created at the same time (after hit soft limit)
+	Max_num_sandbox_creating int `json:"max_num_sandbox_creating"`
+
 	// sandbox options
 	// worker directory, which contains handler code, pid file, logs, etc.
 	Worker_dir string `json:"worker_dir"`
@@ -125,6 +132,20 @@ func (c *Config) Defaults() error {
 
 	if c.Registry_dir == "" {
 		return fmt.Errorf("must specify local registry directory")
+	}
+
+	// mem usage limits for creating sandboxes
+	if c.Create_sandbox_soft_limit == 0.0 {
+		c.Create_sandbox_soft_limit = 0.5
+	} 
+
+	if c.Create_sandbox_hard_limit == 0.0 {
+		c.Create_sandbox_hard_limit = 0.9
+	} 
+
+	// max number of sandboxes being created at the same time (after hit soft limit)
+	if c.Max_num_sandbox_creating == 0 {
+		c.Max_num_sandbox_creating = 4
 	}
 
 	if !path.IsAbs(c.Registry_dir) {
--- a/worker/config/config.go
+++ b/worker/config/config.go
@@ -46,12 +46,11 @@ type Config struct {
 	Handler_cache_size int `json:"handler_cache_size"` //kb
 	Import_cache_size  int `json:"import_cache_size"`  //kb
 
-	// mem usage limits for creating sandboxes
-	Create_sandbox_soft_limit float64 `json:"create_sandbox_soft_limit"` //mem usage percentage, from 0.0 (0%) to 1.0 (100%)
-	Create_sandbox_hard_limit float64 `json:"create_sandbox_hard_limit"` //mem usage percentage, from 0.0 (0%) to 1.0 (100%)
+	// average memory usage for each handler
+	Average_handler_memory_usage int `json:"average_handler_memory_usage"` //kb
 
-	// max number of sandboxes being created at the same time (after hit soft limit)
-	Max_num_sandbox_creating int `json:"max_num_sandbox_creating"`
+	// mem usage limit for creating sandboxes
+	Create_sandbox_hard_limit int `json:"create_sandbox_hard_limit"` //kb
 
 	// sandbox options
 	// worker directory, which contains handler code, pid file, logs, etc.
@@ -132,20 +131,6 @@ func (c *Config) Defaults() error {
 
 	if c.Registry_dir == "" {
 		return fmt.Errorf("must specify local registry directory")
-	}
-
-	// mem usage limits for creating sandboxes
-	if c.Create_sandbox_soft_limit == 0.0 {
-		c.Create_sandbox_soft_limit = 0.8
-	} 
-
-	if c.Create_sandbox_hard_limit == 0.0 {
-		c.Create_sandbox_hard_limit = 0.9
-	} 
-
-	// max number of sandboxes being created at the same time (after hit soft limit)
-	if c.Max_num_sandbox_creating == 0 {
-		c.Max_num_sandbox_creating = 4
 	}
 
 	if !path.IsAbs(c.Registry_dir) {
--- a/worker/config/config.go
+++ b/worker/config/config.go
@@ -46,11 +46,14 @@ type Config struct {
 	Handler_cache_size int `json:"handler_cache_size"` //kb
 	Import_cache_size  int `json:"import_cache_size"`  //kb
 
-	// average memory usage for each handler
-	Average_handler_memory_usage int `json:"average_handler_memory_usage"` //kb
-
 	// mem usage limit for creating sandboxes
-	Create_sandbox_hard_limit int `json:"create_sandbox_hard_limit"` //kb
+	Sandbox_create_hard_limit int `json:"sandbox_create_hard_limit"` //kb
+
+	// size limit of the queue containing handlers waiting to create sandbox
+	Sandbox_create_queue_size int `json:"sanbox_create_queue_size"`
+
+	// time interval to update average handler memory usage
+	Handler_usage_update_interval int `json:"handler_usage_update_interval"` // seconds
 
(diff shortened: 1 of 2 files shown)


worker/config/config.go
```diff
@@ -46,6 +46,12 @@ type Config struct {
 	Handler_cache_size int `json:"handler_cache_size"` //kb
 	Import_cache_size  int `json:"import_cache_size"`  //kb
 
+	// average memory usage for each handler
+	Average_handler_memory_usage int `json:"average_handler_memory_usage"` //kb
```
tylerharter: Let's get rid of this as an option.  The dynamic calculation you have seems better, and I don't think the ability to override that with a constant is worth the complexity.

worker/handler/handler.go
```diff
@@ -106,24 +113,72 @@ func NewHandlerManagerSet(opts *config.Config) (hms *HandlerManagerSet, err erro
...
+		sbHardLimit:        opts.Create_sandbox_hard_limit * 1024,
+		numSbCreating:      0,
 	}
 
+	hms.sbCreatingChan = make(chan bool, 1)
+
 	hms.lru = NewHandlerLRU(hms, opts.Handler_cache_size) //kb
 
+	if hms.avgHandlerMemUsage == 0 {
+		go func() {
```
tylerharter: There's too much complexity here to have it be an anonymous function.  Let's move it out to it's own function.

worker/handler/handler.go
```diff
@@ -106,24 +113,72 @@ func NewHandlerManagerSet(opts *config.Config) (hms *HandlerManagerSet, err erro
...
+						}
+						numHandler++
+						totalMemUsage += handlerUsage(h)
+					}
+				}
+				hms.mutex.Unlock()
+				if numHandler != 0 {
+					hms.avgHandlerMemUsage = totalMemUsage / numHandler
+				}
+				time.Sleep(time.Second)
```
tylerharter: One second seems often.  Can you make it longer (say 10 seconds)?  Also, leave a comment that we should eventually make it configurable.

worker/handler/handler.go
```diff
@@ -106,24 +113,72 @@ func NewHandlerManagerSet(opts *config.Config) (hms *HandlerManagerSet, err erro
...
+	hms.sbCreatingChan = make(chan bool, 1)
+
 	hms.lru = NewHandlerLRU(hms, opts.Handler_cache_size) //kb
 
+	if hms.avgHandlerMemUsage == 0 {
+		go func() {
+			for true {
+				numHandler := 0
+				totalMemUsage := 0
+				hms.mutex.Lock()
```
tylerharter: I'm scared that we're doing a procfs read for every handler while holding a global lock.  Please add some code to measure the latency of this section, and write a WARNING to the log if it takes more than, say, 10ms.

We don't need to prematurely optimize, but I don't want to forget about this when it likely becomes an issue for higher container densities.

worker/handler/handler.go
```diff
@@ -277,6 +338,21 @@ func (h *Handler) RunStart() (ch *sb.Channel, err error) {
 
 	// create sandbox if needed
 	if h.sandbox == nil {
+		// if estimated mem usage exceeds the hard limit, we would block
```
tylerharter: This model feels wrong to me (everybody Get invocation is effectively doing its own polling, and which get to proceed when resources become free is arbitrary).

When we're waiting to proceed, shouldn't we Get be added to a queue, and then a scheduler (implemented by a separate go routine) decides which Get calls to unblock, and when?

worker/handler/handler.go
```diff
@@ -277,6 +338,21 @@ func (h *Handler) RunStart() (ch *sb.Channel, err error) {
...
 	// create sandbox if needed
 	if h.sandbox == nil {
+		// if estimated mem usage exceeds the hard limit, we would block 
+		// all the creation requests until there are more free mem.
+		hms.sbCreatingLock.Lock()
+		for hms.sbHardLimit != 0 && (getMemUsage() + (hms.numSbCreating + 1) * hms.avgHandlerMemUsage > hms.sbHardLimit) {
+			hms.sbCreatingLock.Unlock()
+			select {
+			case <-hms.sbCreatingChan:
+			case <-time.After(4 * time.Second):
```
tylerharter: 4 seconds is an eternity...

worker/handler/handler.go
```diff
@@ -432,6 +509,13 @@ func (h *Handler) nuke() {
 	if err := h.sandbox.Remove(); err != nil {
 		log.Printf("failed to remove sandbox :: %v", err.Error())
 	}
+
+	// a sandbox has been destroyed, attempts to wake up blocked
+	// sandbox creation requests, if any.
+	hms := h.hm.hms
+	if getMemUsage() < hms.sbHardLimit {
```
tylerharter: Since we have to poll (since there is no other way to be alerted when a process relinquishes memory), this kind of signaling is an optimization, right?  As such, have we done perf experiments justifying this optimization?

worker/config/config.go
gtbai: Updated in the new commit.

worker/handler/handler.go
gtbai: Updated in the new commit.

worker/handler/handler.go
gtbai: Updated in the new commit, made this configurable and default value be 10 sec.

worker/handler/handler.go
gtbai: Updated in the new commit.

worker/handler/handler.go
gtbai: That makes sense, updated in the new commit.

worker/handler/handler.go
gtbai: This is replaced by polling in the separate scheduling goroutine in the new commit.

worker/handler/handler.go
gtbai: Got rid of this in the new commit.

worker/handler/handler.go
```diff
@@ -277,6 +360,20 @@ func (h *Handler) RunStart() (ch *sb.Channel, err error) {
 
 	// create sandbox if needed
 	if h.sandbox == nil {
+		// if estimated mem usage exceeds the hard limit, the handler would queue itself
+        hms.sbCreatingLock.Lock()
+        for hms.sbHardLimit != 0 && (getMemUsage() + (hms.numSbCreating + 1) * hms.avgHandlerMemUsage > hms.sbHardLimit) {
```
tylerharter: I don't understand the purpose of having both this and the scheduler that decides when to release blocked RunStarts.

worker/handler/handler.go
```diff
@@ -277,6 +360,20 @@ func (h *Handler) RunStart() (ch *sb.Channel, err error) {
...
+            hms.sbCreatingLock.Unlock()
+			if h.sbCreatingChan == nil {
+				h.sbCreatingChan = make(chan bool, 1)
+			}
+			hms.sbCreatingQueue <- h
+			<-h.sbCreatingChan // waiting for wake-up signal
+            hms.sbCreatingLock.Lock()
```
tylerharter: Weird indenting in a few places

worker/handler/handler.go
```diff
@@ -277,6 +360,20 @@ func (h *Handler) RunStart() (ch *sb.Channel, err error) {
 
 	// create sandbox if needed
 	if h.sandbox == nil {
+		// if estimated mem usage exceeds the hard limit, the handler would queue itself
+        hms.sbCreatingLock.Lock()
```
tylerharter: Do we need this since we have a central scheduler, which we communicate with via the chan?

worker/handler/handler.go
```diff
@@ -106,24 +115,92 @@ func NewHandlerManagerSet(opts *config.Config) (hms *HandlerManagerSet, err erro
...
+		}
+		hms.mutex.Unlock()
+		if numHandler != 0 {
+			hms.avgHandlerMemUsage = totalMemUsage / numHandler
+		}
+		time.Sleep(time.Duration(updateInterval) * time.Second)
+	}
+}
+
+func (hms *HandlerManagerSet) scheduleSbCreatingReqs() {
```
tylerharter: I think this is important enough to warrant some commenting

worker/handler/handler.go
gtbai: Sorry for the confusion about this, will fix in the next commit.

worker/handler/handler.go
gtbai: Removed in the new commit.

worker/handler/handler.go
gtbai: Removed in the new commit.

worker/handler/handler.go
gtbai: Added comments in the new commit.

tylerharter:
Hey Guangtong, I've been thinking a lot about how we can guarantee we never go over the memory limits.  Unless we're integrating closely with cgroup mem limits, I don't think we can ever get this working perfectly.  If you're interested in still working on this, let's chat in person sometime about the design.

PR #55, opened 2019-04-25
[Vendor Only] Work of Lambda Team in S19 CS838 Edge Computing Lab

gtbai:
This is the vendor part of PR https://github.com/open-lambda/open-lambda/pull/53.

> In this PR, we added the following two features for workers:
> 
> 1. Enable requests of the same lambda to reuse a handler.
>    
>    * To accomplish this, we moved the code that adds a new handler to the `HandlerManagerSet .HandlerManager.handlers` list.
> 2. Added request queuing mechanism (mem usage hard limit) for workers.
>    
>    * Added a utility function to get the current machine memory usage, making use of [shirou/gopsutil](https://github.com/shirou/gopsutil)
>    * Added both a soft limit (`Config.Create_sandbox_soft_limit `/`HandlerManagerSet.sbSoftLimit`) and a hard limit (`Config.Create_sandbox_hard_limit `/`HandlerManagerSet.sbHardLimit`) for memory usage when creating new sandboxes.
>      
>      * When memory usage exceeds the soft limit, at most `Config.Max_num_sandbox_creating `/`HandlerManagerSet.maxNumSbCreating ` sandboxes can be being created at the same time.
>      * When memory usage exceeds the hard limit, requests creating new sandboxes will be blocked, until memory usage decreases.


tylerharter:
Why are there 89 files?

gtbai:
> Why are there 89 files?

Sorry for the confusion. I mistakenly re-vendored all the packages in the previous PR. In this PR I only added [shirou/gopsutil](https://github.com/shirou/gopsutil) (which we used to get the current mem usage of the host) and its dependencies, so less files.

PR #82, opened 2019-06-30
Adjust cgroup mem limit in Pause and Unpause

gtbai:
Adjust cgroup memory limit after a container is paused or unpaused.

### DIFF
--- a/src/sandbox/cgroupFactory.go
+++ b/src/sandbox/cgroupFactory.go
@@ -18,368 +18,367 @@ var cgroupList []string = []string{
 	"blkio", "cpu", "devices", "freezer", "hugetlb",
 	"memory", "perf_event", "systemd", "pids"}
 
-	// if there are fewer than CGROUP_RESERVE available, more will be created.
-	// If there are more than 2*CGROUP_RESERVE available, they'll be released.
-	const CGROUP_RESERVE = 16
+// if there are fewer than CGROUP_RESERVE available, more will be created.
+// If there are more than 2*CGROUP_RESERVE available, they'll be released.
+const CGROUP_RESERVE = 16
 
-	type Cgroup struct {
-		Name string
-		pool *CgroupPool
+type Cgroup struct {
+	Name string
+	pool *CgroupPool
+}
+
+type CgroupPool struct {
+	Name     string
+	memPool  *MemPool
+	ready    chan *Cgroup
+	recycled chan *Cgroup
+	quit     chan chan bool
+	nextId   int
+}
+
+func NewCgroupPool(name string) (*CgroupPool, error) {
+	pool := &CgroupPool{
+		Name:     path.Base(config.Conf.Worker_dir) + "-" + name,
+		ready:    make(chan *Cgroup, CGROUP_RESERVE),
+		recycled: make(chan *Cgroup, CGROUP_RESERVE),
+		quit:     make(chan chan bool),
+		nextId:   0,
 	}
 
-	type CgroupPool struct {
-		Name     string
-		memPool  *MemPool
-		ready    chan *Cgroup
-		recycled chan *Cgroup
-		quit     chan chan bool
-		nextId   int
-	}
-
-	func NewCgroupPool(name string) (*CgroupPool, error) {
-		pool := &CgroupPool{
-			Name:     path.Base(config.Conf.Worker_dir) + "-" + name,
-			ready:    make(chan *Cgroup, CGROUP_RESERVE),
-			recycled: make(chan *Cgroup, CGROUP_RESERVE),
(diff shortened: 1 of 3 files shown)


src/sandbox/cgroupFactory.go
```diff
@@ -29,6 +29,7 @@ type Cgroup struct {
 
 type CgroupPool struct {
 	Name     string
+	memPool  *MemPool
```
tylerharter: I don't think the MemPool abstraction should bleed across layers.  The cgroup pool should be dumb (i.e., it doesn't do policy), and should just expose an API like this: changeMemLimit(mb int)

The sock.go layer should be the one to decide when to upsize/downsize the memory limit.  I was suggesting changing the Pause/Unpause functions in sock.go, rather than implement that logic here in cgroupFactory.go.

One reason for implementing the mechanism, changeMemLimit(mb int), here and the resizing policy in sock.go is that there are other opportunities to play with the mem limits that we can only do at a higher level.  For example, fork servers are unlikely to grow in memory after they've done they're imports, so we should eventually consider constricting them (paused or not).

src/sandbox/cgroupFactory.go
```diff
@@ -186,6 +186,64 @@ func (cg *Cgroup) setFreezeState(state string) error {
...
+
+// get mem limit in bytes
+func (cg *Cgroup) setMemLimit(newLimit int64) error {
+	limitPath := cg.Path("memory", "memory.limit_in_bytes")
+	// if mem limit cannot be set, return error
+	if err := ioutil.WriteFile(limitPath, []byte(fmt.Sprintf("%d", newLimit)), os.ModeAppend); err != nil {
+		return err
+	}
+
+	timeout := 5 * time.Second
```
tylerharter: Is this instantaneous, based on your experiments or documentation?  I only thought we would need to do this for freezing/unfreezing.

src/sandbox/sock.go
```diff
@@ -159,12 +159,62 @@ func (c *SOCKContainer) populateRoot() (err error) {
...
+	}
+
+	// UNPAUSE STEP 3: unpause the cgroup
+	if err := c.cg.Unpause(); err != nil {
+		return err
+	}
+
+	c.printf("available mem after pause: %d", c.pool.mem.adjustAvailableMB(0))
+	memLimit, err := c.cg.getMemLimit()
+	c.printf("mem limit read from file after pause: %d", memLimit)
```
tylerharter: unpause, right?

src/sandbox/sock.go
```diff
@@ -159,12 +159,62 @@ func (c *SOCKContainer) populateRoot() (err error) {
...
+	if err := c.cg.setMemLimit(newLimit); err != nil {
+		return err
+	}
+
+	// UNPAUSE STEP 3: unpause the cgroup
+	if err := c.cg.Unpause(); err != nil {
+		return err
+	}
+
+	c.printf("available mem after pause: %d", c.pool.mem.adjustAvailableMB(0))
```
tylerharter: This is a little expensive, because we need to send a request to the memPool over a chan (it's like a server).  Your earlier call to adjustAvailableMB already returns the available mem, so just save it there if you want to do this print

src/sandbox/cgroupFactory.go
```diff
@@ -186,6 +186,64 @@ func (cg *Cgroup) setFreezeState(state string) error {
...
+
+	start := time.Now()
+	for {
+		memLimit, err := cg.getMemLimit()
+		if err != nil {
+			return fmt.Errorf("failed to get mem limit :: %v", err)
+		}
+
+		if memLimit == newLimit {
```
tylerharter: This may never happen, even if we succeed.  Take a look here: https://www.kernel.org/doc/Documentation/cgroup-v1/memory.txt

> successful write to this file does not guarantee a successful setting of
> this limit to the value written into the file. This can be due to a
> number of factors, such as rounding up to page boundaries or the total
> availability of memory on the system. The user is required to re-read
> this file after a write to guarantee the value committed by the kernel.

src/sandbox/cgroupFactory.go
```diff
@@ -186,6 +186,64 @@ func (cg *Cgroup) setFreezeState(state string) error {
 	}
 }
 
+// get mem usage in bytes
+func (cg *Cgroup) getMemUsage() (int64, error) {
```
tylerharter: Would be nice to have the units (e.g., KB, MB) as part of the name of these functions

gtbai:
> Shall I merge (please confirm you've tested)? 
Yes please, we tested by printing out debug messages.

> We should have some tests in test.py too, but that could be a separate PR.
Sure.

PR #90, opened 2019-09-09
Introduce Timeouts for Lambda Functions [WIP]

wchen329:
# Synopsis
This PR introduces optional timeouts for running Lambda functions in OpenLambda. These timeouts can be set by the programmer, or the OpenLambda cluster manager in config.json (lambda_max_timeout). These timeouts are set as integers in units **milliseconds**. lambda_max_timout in config.json is the default timeout and also the maximum possible timeout, while the timeout set by the programmer through #ol-timeout overrides the default (so long it is less than or equal to the default). If both are set, then the lesser of the two is set. To disable default timeouts, set the timeout value in config.json to an integer less than or equal to 0. If a Lambda runs longer than the designated timeout, then it will automatically cancel execution and write a error message to the response sent to the original client.

This PR also aims to refactor the interface between the sandboxes and http handlers.

## Synchronization
There is a mutex used between the Timer goroutine (responsible for timing out a process) and that used by the main thread. This mutex was used in order to prevent potentially dangerous conditions of "double dipping" on the deletes. In langauges with manual resource allocation, this almost certainly results in an abort. Not sure if this actually an issue in Go.

## Testing
Tests are not added yet. But manual testing has been done to show that this in certain situations does work.

## Other Changes
* Fix the Makefile (i.e. add the 'all' dummy target as a dependency to the 'test-all' target [it's generally a good idea to make sure the make is up to date before performing tests])

## Is this Ready?
No, not yet.


src/lambda/lambda.go
```diff
@@ -279,6 +308,25 @@ func parseMeta(codeDir string) (meta *sandbox.SandboxMeta, err error) {
 					imports = append(imports, val)
 				}
 			}
+		} else if parts[0] == "#ol-timeout" {
+			// case: correct amount of parts. Accepts *:* where * is
+			// NOT in {:, space, \n, \r} etc.
+			if len(parts) == 2 {
```
wchen329: Todo: generalize this check to each of the three specs.

src/common/config.go
```diff
@@ -55,6 +55,10 @@ type Config struct {
 	// which OCI implementation to use for the docker sandbox (e.g., runc or runsc)
 	Docker_runtime string `json:"docker_runtime"`
 
+	// The max lambda timeout given in milliseconds
+	// If no timeout is given by the lambda, this max timeout is also the default
+	Lambda_Max_timeout int64 `json:"lambda_max_timeout"`
```
tylerharter: 1. let's put it in LimitsConfig
2. don't need to prefix with "lambda_" (or we would prefix nearly everything like that)
3. let's suffix with "_ms" so it's obvious what the units are (like we suffix with "_mb" below)

wchen329: Sure we can do that! :+1: 

(by the way I probably won't make much further changes until the weekend since I've got to work on a ECE 554 project that's due in a couple days) I might be able to slip this in tomorrow (hopefully) after we're done with the "mini-project")

wchen329:
Since it's been a while since I've worked on this, I'm going to open (1) a new PR on a new branch with the refactoring stiff and then (2) reopen another PR with the timeout functionality added, rebuilding from the canonical master

PR #91, opened 2019-09-09
Build ol library inside the container to avoid host compatibility concerns

NathanTP:
I know the official way to use OpenLambda is on the development VM, but I was trying to run it on Arch Linux anyway. The problem I ran into was that ol.so was built on the host and linked against the host's version of libpython. On the standard build VM this is fine because the container and the host are running the same version of everything, but on different hosts, those dependencies might differ. The solution is to build everything that runs in the container in the container itself to ensure compatibility. I also changed spin.c to link statically for good measure (it wasn't causing me problems, but it might conceivably run into a similar issue to ol.so in the future).

Anyway, I figured I'd PR my fix in case it helped anyone else.

### DIFF
--- a/lambda/Dockerfile
+++ b/lambda/Dockerfile
@@ -8,7 +8,11 @@ RUN pip3 install virtualenv requests tornado==4.5.3
 
 # for SOCK container engine
 COPY sock2.py /
-COPY ol.so /
+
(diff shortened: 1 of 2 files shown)


tylerharter:
Good idea, thanks!

PR #93, opened 2020-01-23
Convert Latency methods to a pointer receiver

NathanTP:
Previously l.Milliseconds would always be 0 becaust lat.T1() worked on a copy of lat. One implication of this is that the autoscaling algorithms would never scale the number of containers (since the algorithm was based on estimated outstanding work, which was always 0).

### DIFF
--- a/src/common/stats.go
+++ b/src/common/stats.go
@@ -99,15 +99,15 @@ type Latency struct {
 }
 
 // record start time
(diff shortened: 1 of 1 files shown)


tylerharter:
Good catch, thanks!  I didn't notice since it was still logging the latencies fine.

PR #96, opened 2020-02-27
Enhance ol command

### LINKED ISSUE #95: make test all failed - "using existing OL directory"
Hi to all
i  have  ran make test-all after installed open lambda successfully
but just 1 test passed and 17 tests failed and when i was checking errors i encountered a same error for all failed tests with following content:
""using existing OL directory at /usr/local/open-lambda/test-dir2"
"previous worker may be running, /usr/local/open-lambda/test-dir2/worker/worker.pid already exists"

hot machine : ubuntu 18
docker version :
18.09.7
go version:
go version go1.13.8 linux/amd64
i installed ubuntu along side windows 10

can any one help me  to fix these errors?

thanks.

GindaChen:
Fixes #95.
The following enhancement are made:
1. Add kill progress checking. `ol kill` will show the progress of killing the worker process
2. Add async kill option. `ol kill -a` will send signal to the worker process and exit immediately, showing the progress. `ol status`


GindaChen:
@tylerharter I will raise a second PR about the kill. My implementation is a bit tricky, and I'm trying to make it more readable...

GindaChen:
> I don't think we should have an async kill as part of the tool. Somebody could always run the command in the background if they wanted, right?

I see the point. I will revert the change.

> I like that we can see the progress. Can we make it so that we don't timeout during kill as long as progress is being made?

Sure.

GindaChen:
@tylerharter The make test has not passed after the change. Do we run test when submitting the PR (like circle ci?)

GindaChen:
The tests has passed on cloudlab.

src/common/config.go
```diff
@@ -218,6 +218,12 @@ func checkConf() error {
 		return fmt.Errorf("Unknown Sandbox type '%s'", Conf.Sandbox)
 	}
 
+	// check proc number limit. if proc num == 1, then any fork will fail.
+	// TODO: (Discussion) what is the proc num we shall set as a hard lower bound?
```
GindaChen: @tylerharter I think we should add a proc limit here to safe-check the config. Apperantly, `proc<10` means the system is basically useless (it will crash after 3~4 plain requests). But whether we shall set this here is worth discussion.

src/main.go
```diff
@@ -33,12 +33,150 @@ func getOlPath(ctx *cli.Context) (string, error) {
 	return filepath.Abs(olPath)
 }
 
+func getOlWorkerProc(ctx *cli.Context) (int, error) {
```
GindaChen: I have a generic question about the generic environment / configuration of open lambda. 

**`ol` command only work in the "right" directory**
I think we assume open lambda worker will be a singleton in the system, and therefore both the `default-ol` directory and config files should be fixed at some mount point. But the command itself is set to be a relative path, and only take effect if the sys admin is in the correct directory. Why is this the case?

src/main.go
```diff
@@ -33,12 +33,150 @@ func getOlPath(ctx *cli.Context) (string, error) {
...
+	//	err:
+	//		Error message of all kind.
+
+	// Assumption:
+	//	1. At most one ol worker is running in the system.
+	//  2. This process is the only other ol admin process that could be running.
+	// TODO: Assumption 2 is too strong as there could be many `ol status` running.
+
+	// Get the running pid whose program name is `ol` (at least one - this one)
+	out, err := exec.Command("ps", "-C", "ol", "-o", "pid=").Output()
```
GindaChen: Problem here:
1. We assume the executable will be named `ol`. This hard-code value is probably not a good idea, and we shall specify the path to the `ol` command.
2. There could be multiple `ol` running at the same time (multiple admins doing status check, etc). To discern the cases we need more robust mechanism to do this

GindaChen:
DEBUG: sandbox created / deleted

PR #101, opened 2022-02-11
S22

emmabrostrom:
Updated to Ubuntu 20.04 and Python 3.8, also updated numpy install testers to work with these higher versions.

### DIFF
--- a/boss-test.py
+++ b/boss-test.py
@@ -1,0 +1,117 @@
+import requests
+import os, sys
+from subprocess import run
+import time
+
+api_key = None
+boss_port = None
+
+def read_json(path):
+    with open(path) as f:
+        return json.load(f)
+
+def write_json(data, path):
+    with open(path, "w") as f:
+        return json.dump(data, f, indent=2)
+
+def boss_get(resource, check=True):
+    url = f"http://localhost:{boss_port}/{resource}"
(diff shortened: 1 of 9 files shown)


boss-test.py
```diff
@@ -0,0 +1,117 @@
+import requests
```
tylerharter, line 1: Let's leave this file out of the PR since we won't be pushing boss code to the master branch for a while.

test.py
```diff
@@ -470,7 +480,7 @@ def tests():
         stress_one_lambda(procs=8, seconds=15)
 
     with TestConf(features={"reuse_cgroups": True}):
-        call_each_once(lambda_count=100, alloc_mb=1)
+        call_each_once(lambda_count=100, alloc_mb=10)
```
tylerharter, line 483: Is this change intentional/necessary?

tylerharter (changes requested):
This looks great, thanks for updating Python/Tornado/Ubuntu in the SOCK containers!  Just some minor comments.

tylerharter:
Merged it in!  I found a way in GitHub to add commits on top of the ones in a PR before merging, so I was able to directly delete the sec.c and boss-test.py.  Thanks for upgrading Python/Ubuntu/Tornado!

PR #134, opened 2022-11-09
Seccomp option

kwangjong:
Added an argument in config file for enabling/disabling seccomp filtering

### DIFF
--- a/lambda/runtimes/python/server.py
+++ b/lambda/runtimes/python/server.py
@@ -166,15 +166,18 @@ def main():
     global bootstrap_path
 
     if len(sys.argv) < 2:
-        print("Expected execution: chroot <path_to_root_fs> python3 server.py <path_to_bootstrap.py> [cgroup-count]")
+        print("Expected execution: chroot <path_to_root_fs> python3 server.py <path_to_bootstrap.py> [cgroup-count] [enable-seccomp]")
         print("    cgroup-count: number of FDs (starting at 3) that refer to /sys/fs/cgroup/..../cgroup.procs files")
+        print("    enable-seccomp: true/false to enable or disables seccomp filtering")
         sys.exit(1)
 
     print('server.py: started new process with args: ' + " ".join(sys.argv))
 
-    return_code = ol.enable_seccomp()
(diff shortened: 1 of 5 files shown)


test-registry/syslog/f.py
```diff
@@ -0,0 +1,6 @@
+import subprocess
+from ctypes import *
+
+def f(event):
+	libc = CDLL("/lib/x86_64-linux-gnu/libc.so.6")
+	return libc.syscall(103, 0, "")
```
tylerharter, line 6: is there a constant somewhere we can use instead of hardcoding 103?

lambda/runtimes/python/server.py
```diff
@@ -166,15 +166,18 @@ def main():
...
         print("    cgroup-count: number of FDs (starting at 3) that refer to /sys/fs/cgroup/..../cgroup.procs files")
+        print("    enable-seccomp: true/false to enable or disables seccomp filtering")
         sys.exit(1)
 
     print('server.py: started new process with args: ' + " ".join(sys.argv))
 
-    return_code = ol.enable_seccomp()
-    assert return_code >= 0
-    print('seccomp enabled')
+    #enable_seccomp if enable-seccomp is not passed
```
tylerharter, line 176: can we pass in a dict of options containing a seccomp option?  Then we can pass in other configs in the same way

src/common/config.go
```diff
@@ -74,6 +74,7 @@ type FeaturesConfig struct {
 	Reuse_cgroups       bool `json:"reuse_cgroups"`
 	Import_cache        bool `json:"import_cache"`
 	Downsize_paused_mem bool `json:"downsize_paused_mem"`
+	Enable_seccomp      bool `json:"enable_seccomp"`
```
tylerharter, line 77: should we have a sub structure with options that get passed directly to the container?

tylerharter (changes requested):

tylerharter:
Should we make this part of the continuous integration tests?  https://github.com/open-lambda/open-lambda/blob/master/.github/workflows/ci.yml

PR #158, opened 2023-05-15
Add CPU profiler for server

NataRich:
Updated `main.go` and `server.go` files for adding the CPU profiling feature.

Considering that users might profile the program for an arbitrarily long time intentionally or by mistake, I manually set the max profiling duration to 10 minutes (600 seconds) in both files to avoid holding the thread for too long. However, this is not the optimal solution. We may consider including this parameter in the configuration file.


tylerharter:
Good work!  This is implementation is fairly similar to the memory profiler, which is good from the perspective of users learning to use both.  I think it needs to be more different, though, because memory profiling is a point-in-time measurement, whereas CPU profiling is measuring a range.

I think a common use case will be start profiling, run a benchmark, then stop profiling.  You might not know how long the benchmark will run, so users shouldn't need to specify that.  Instead, I think it would be good to have separate "ol pprof cpu-start" and "ol pprof cpu-stop" commands.  The cpu-start command should print a helpful messages about running cpu-stop when finished, and cpu-stop should print the messaged about how to view the profile.

src/server/server.go
```diff
@@ -63,6 +65,23 @@ func PprofMem(w http.ResponseWriter, r *http.Request) {
         }
 }
 
+func PprofCpu(w http.ResponseWriter, r *http.Request) {
+	runtime.GC()
```
tylerharter, line 69: Do we need garbage collection before CPU profiling?  I'm guessing not.

src/server/server.go
```diff
@@ -63,6 +69,32 @@ func PprofMem(w http.ResponseWriter, r *http.Request) {
...
+	w.Header().Add("Content-Type", "application/octet-stream")
+
+	if err := pprof.StartCPUProfile(w); err != nil {
+	    log.Fatal("could not start CPU profile: ", err)
+	}
+
+	select {
+	case <-cpuProfTimeout:
+		break
+	case <-time.After((1 << 63 - 1) * time.Nanosecond): // max duration
```
tylerharter, line 83: Is there value in having a 100+ year timeout?

src/server/server.go
```diff
@@ -63,6 +69,32 @@ func PprofMem(w http.ResponseWriter, r *http.Request) {
...
+func PprofCpuStart(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Starting cpu profiling\n")
+	w.Header().Add("Content-Type", "application/octet-stream")
+
+	if err := pprof.StartCPUProfile(w); err != nil {
+	    log.Fatal("could not start CPU profile: ", err)
+	}
+
+	select {
+	case <-cpuProfTimeout:
```
tylerharter, line 81: Does this end point need to block until profiling is completed, or can we just send back a message immediately to indicate that profiling has started?

src/server/server.go
```diff
@@ -63,6 +69,32 @@ func PprofMem(w http.ResponseWriter, r *http.Request) {
         }
 }
 
+func PprofCpuStart(w http.ResponseWriter, r *http.Request) {
+	log.Printf("Starting cpu profiling\n")
+	w.Header().Add("Content-Type", "application/octet-stream")
+
+	if err := pprof.StartCPUProfile(w); err != nil {
```
tylerharter, line 76: Instead of writing the profile info back to w, I would write it to a temp file and just send back a confirmation it is started.  I think users will want the profiling data after they call stop, so the stop call should (1) stop the profiling (2) read the temp file and (3) send the profile data back.

tylerharter (changes requested):
Some cases I'm hoping you can think about:
1. if somebody starts profiling while it is already started, should we try to do it twice (overlapping), or return an error message?  I'm guessing return an error, but I'm interested in your thoughts.
2. if somebody tries to stop profiling when it's not running, should we return an error, or is there any other reasonable behavior?
3. if somebody starting profiling, then kills the worker without ever calling stop, what should we do?  Nothing, or create a profile somewhere?

Also, there are some linting issues to fix: https://github.com/open-lambda/open-lambda/pull/158/checks

src/server/server.go
NataRich: I meant to write a wait-and-interrupt mechanism. I chose a 100+ year timeout (actually, the max duration the library supports) to simulate the "infinite wait" until an "interrupt" (`cpuProfTimeout` channel gets filled). [Reference](https://stackoverflow.com/a/23331563/11247758)

src/server/server.go
NataRich: Yes, I think this makes more sense. I will use a temp file :)

src/server/server.go
tylerharter: If it's infinite wait, then you don't need the select/case.  It's just a one-liner:

<-cpuProfTimeout

But I'm hoping we can get rid of the channel altogether.

tylerharter:
Looking forward to the updated PR!

I agree with your thinking about when OL should panic/exit -- definitely not due to user error.

"I do not know if there is a way to determine whether the CPU profile has been started" => if you're writing at a temp file after cpu start, then one way is to check whether it exists (assumes you delete the temp file after cpu stop).

src/server/server.go
```diff
@@ -59,10 +65,64 @@ func PprofMem(w http.ResponseWriter, r *http.Request) {
...
+	pprof.StopCPUProfile()
+	tempFilename := cpuTemp.Name()
+	cpuTemp.Close()
+	cpuTemp = nil
+	defer os.Remove(tempFilename) // deferred cleanup
+
+	log.Printf("Reading from %s\n", tempFilename)
+	buffer, err := ioutil.ReadFile(tempFilename)
+	if err != nil {
+		log.Fatal("could not read file: ", err)
```
tylerharter, line 116: let's not have any log.Fatals or panics.  Just sending an error back for the user to see is right for this.

src/server/server.go
```diff
@@ -24,8 +24,14 @@ const (
 	STATS_PATH  = "/stats"
 	DEBUG_PATH  = "/debug"
 	PPROF_MEM_PATH  = "/pprof/mem"
+	PPROF_CPU_START_PATH = "/pprof/cpu-start"
+	PPROF_CPU_STOP_PATH = "/pprof/cpu-stop" 
 )
 
+// temporary file storing cpu profiled data
```
tylerharter, line 31: should these be protected by a lock?

src/server/server.go
```diff
@@ -59,10 +65,64 @@ func PprofMem(w http.ResponseWriter, r *http.Request) {
...
-            log.Fatal("could not write memory profile: ", err)
+	    	log.Fatal("could not write memory profile: ", err)
         }
 }
 
+// Starts CPU profiling
+func PprofCpuStart(w http.ResponseWriter, r *http.Request) {
+	// user error: double start (previous profiling not stopped yet)
+	if cpuTemp != nil {
+		log.Printf("Already started cpu profiling\n")
```
tylerharter, line 76: why not write this back to w along with the status code?  Probably more helpful to the user there than buried in the log output.

src/server/server.go
```diff
@@ -59,10 +65,64 @@ func PprofMem(w http.ResponseWriter, r *http.Request) {
         runtime.GC()
 	w.Header().Add("Content-Type", "application/octet-stream")
         if err := pprof.WriteHeapProfile(w); err != nil {
-            log.Fatal("could not write memory profile: ", err)
+	    	log.Fatal("could not write memory profile: ", err)
         }
 }
 
+// Starts CPU profiling
+func PprofCpuStart(w http.ResponseWriter, r *http.Request) {
```
tylerharter, line 73: The way I would probably do this:
1. have a separate function that starts profiling OR returns an error
2. this function will call the helper.   If there is an error, write back whatever it is with a status code.  Then you can handle all the errors in one place.

tylerharter (changes requested):
Getting very close!

src/server/server.go
NataRich: Yes, I think `cpuTempFile` should be protected by a lock, and this would not result in any loss in the performance of interest. In extreme cases, without a lock, it is possible for resource contention to occur - threads 1 & 2 enter `PprofCpuStart` at the same time (both passing the `nil` check) and create two temp files; then, thread 1 modifies `cpuTempFile` and `StartCPUProfile()` smoothly, but thread 2 then modifies it, too, and fails to `StartCPUProfile()`, in which case the value stored in `cpuTempFile` will never be valid, and the file to which the data are actually written (by thread 1) cannot be tracked. So, I added a mutex lock to avoid this in the latest update.

tylerharter (approved):
Looks great, thanks for building this and for the revisions!

PR #177, opened 2023-06-09
S23 azure

keting-ai:
The azure part of s23 works


src/bench/bench.go
```diff
@@ -128,9 +128,9 @@ func run_benchmark(ctx *cli.Context, name string, tasks int, functions int, func
 		waiting -= 1
 	}
 
-	if errors > (errors+successes)/100 {
-		panic(fmt.Sprintf(">1%% of requests failed (%d/%d)", errors, errors+successes))
-	}
+	// if errors > (errors+successes)/100 {
```
tylerharter, line 131: Let's not get rid of this safety check.  You could always make the threshold configurable if you really want.

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+	privateAddr  string
+	publicAddr   string
+}
+
+func NewAzureWorkerPool() *WorkerPool {
+	conf, err := ReadAzureConfig()
+	if err != nil {
+		log.Fatalln(err)
+	}
+	num := conf.Resource_groups.Rgroup[0].Numvm
```
tylerharter, line 39: Do we support multiple resource groups?  If not, we should probably check the length and return an error if it is not 1.

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+	nicName      string
+	publicIPName string
+	privateAddr  string
+	publicAddr   string
+}
+
+func NewAzureWorkerPool() *WorkerPool {
+	conf, err := ReadAzureConfig()
+	if err != nil {
+		log.Fatalln(err)
```
tylerharter, line 37: Probably better to return an error instead of panic'ing

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	num := conf.Resource_groups.Rgroup[0].Numvm
+	workers := make(map[string]*AzureWorker, num)
+	pool := &AzureWorkerPool{
+		workerNum: num,
+		workers:   &workers,
+		nextId:    num + 1,
+	}
+	for i := 0; i < num; i++ {
+		worker_i := new(AzureWorker)
```
tylerharter, line 47: Instead of initializing one by one, let's use this:

worker_i := &AzureWorker{
    pool = ...
}

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+		worker_i.pool = pool
+		worker_i.privateAddr = *conf.Resource_groups.Rgroup[0].Vms[i].Net_ifc.Properties.IPConfigurations[0].Properties.PrivateIPAddress
```
tylerharter, line 49: better to pull the VM into a variable than repeat the same long expression:

vm := conf.Resource_groups.Rgroup[0].Vms[i]

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// TODO: make AzureCreateVM multiple-threaded
+func (pool *AzureWorkerPool) CreateInstance(worker *Worker) error {
+	log.Printf("creating an azure worker\n")
+	conf, err := AzureCreateVM(worker)
+	if err != nil {
+		return err
+	}
+	var private string
```
tylerharter, line 81: Why not use the := syntax below (like you do for all the other variables)?

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+	pool.nextId = pool.workerNum + 1
+
+	(*pool.workers)[azworker.workerId] = azworker
+	worker.workerId = azworker.workerId
+	worker.workerIp = azworker.privateAddr
+
+	return nil
+}
+
+func (worker *Worker) startWorker() {
```
tylerharter, line 117: The the type is worker, then start() is a better function name than startWorker() -- the latter is a bit redundant

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+		sshcmd := exec.Command("ssh", "-i", "~/.ssh/ol-boss_key.pem", user.Username+"@"+worker.workerIp, "-o", "StrictHostKeyChecking=no", "-C", cmd)
+		stdoutStderr, err := sshcmd.CombinedOutput()
+		fmt.Printf("%s\n", stdoutStderr)
+		if err == nil {
+			break
+		}
+		tries -= 1
+		if tries == 0 {
+			fmt.Println(sshcmd.String())
+			panic(err)
```
tylerharter, line 144: Why do we want to panic instead of returning an error?

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+	user, err := user.Current()
+	if err != nil {
+		panic(err)
+	}
+	cmd := fmt.Sprintf("cd %s; %s", cwd, "sudo ./ol worker down")
+	log.Printf("Try to ssh into the worker and kill the process")
+	tries := 10
+	for tries > 0 {
+		log.Printf("debug: %s\n", worker.privateAddr)
+		sshcmd := exec.Command("ssh", "-i", "~/.ssh/ol-boss_key.pem", user.Username+"@"+worker.privateAddr, "-o", "StrictHostKeyChecking=no", "-C", cmd)
```
tylerharter, line 164: Should SSH key name be a config option?

Shouldn't the SSH stuff be part of some general code that is shared between the different platforms (GCP, Azure, etc)?

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+		stdoutStderr, err := sshcmd.CombinedOutput()
+		fmt.Printf("%s\n", stdoutStderr)
+		if err == nil {
+			break
+		}
+		tries -= 1
+		if tries == 0 {
+			fmt.Println(sshcmd.String())
+			panic(err)
```
tylerharter, line 173: Let's avoid panic whenever possible

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+		log.Printf("debug: %s\n", worker.privateAddr)
+		sshcmd := exec.Command("ssh", "-i", "~/.ssh/ol-boss_key.pem", user.Username+"@"+worker.privateAddr, "-o", "StrictHostKeyChecking=no", "-C", cmd)
+		stdoutStderr, err := sshcmd.CombinedOutput()
+		fmt.Printf("%s\n", stdoutStderr)
+		if err == nil {
+			break
+		}
+		tries -= 1
+		if tries == 0 {
+			fmt.Println(sshcmd.String())
```
tylerharter, line 172: Let's use log.Printf instead of fmt in general.

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+	cleanupVM(worker)
+
+	// shrink length
+	conf_lock.Lock()
+	conf, _ := ReadAzureConfig()
+	conf.Resource_groups.Rgroup[0].Numvm -= 1
+	// shrink slice
+	conf.Resource_groups.Rgroup[0].Vms[worker.configPosit] = conf.Resource_groups.Rgroup[0].Vms[len(conf.Resource_groups.Rgroup[0].Vms)-1]
+	conf.Resource_groups.Rgroup[0].Vms = conf.Resource_groups.Rgroup[0].Vms[:conf.Resource_groups.Rgroup[0].Numvm]
+	//fmt.Println(*conf.Resource_groups.Rgroup[0].Vms[worker.configPosit].Vm.Name)
```
tylerharter, line 196: Let's avoid leaving in commented out code

src/boss/azure_worker.go
```diff
@@ -0,0 +1,214 @@
...
+		// if all workers has been deleted, don't do this
+		// if the worker to be deleted is at the end of the list, don't do this
+
+		//TODO: fix this..?
+		(*worker.pool.workers)[*conf.Resource_groups.Rgroup[0].Vms[worker.configPosit].Vm.Name].configPosit = worker.configPosit
+	}
+	worker.pool.workerNum -= 1
+	WriteAzureConfig(conf)
+	log.Printf("Deleted the worker and worker VM successfully\n")
+	conf_lock.Unlock()
```
tylerharter, line 207: Let's use the defer unlock pattern (instead of unlocking at the end)

src/boss/cloudvm/azure.go
```diff
@@ -112,6 +114,14 @@ func randomString() string {
 	return strconv.Itoa(r.Int())
 }
 
+func AzureCreateVM(worker *Worker) (*AzureConfig, error) {
+	subscriptionId = os.Getenv("AZURE_SUBSCRIPTION_ID")
+	if len(subscriptionId) == 0 {
+		log.Fatal("AZURE_SUBSCRIPTION_ID is not set.")
```
tylerharter, line 120: Why not return an error instead?

src/boss/cloudvm/worker.go
```diff
@@ -22,6 +22,14 @@ func NewWorkerPool(platform string, worker_cap int) (*WorkerPool, error) {
 	taskLog.SetFlags(log.Lmicroseconds)
 
 	var pool *WorkerPool
+<<<<<<< HEAD:src/boss/worker.go
```
tylerharter, line 25: Does this even compile?

src/common/config.go
```diff
@@ -150,7 +150,7 @@ func LoadDefaults(olPath string) error {
 	Conf = &Config{
 		Worker_dir:        workerDir,
 		Server_mode:       "lambda",
-		Worker_url:        "localhost",
+		Worker_url:        "0.0.0.0",
```
tylerharter, line 153: Let's leave it as localhost (should go with the safer default).

src/boss/azure_vm.go
```diff
@@ -0,0 +1,889 @@
...
+
+func createVirtualMachine(ctx context.Context, cred azcore.TokenCredential, networkInterfaceID string, new_diskID string, newDiskName string, vmName string) (*armcompute.VirtualMachine, error) {
+	vmClient, err := armcompute.NewVirtualMachinesClient(subscriptionId, cred, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	//require ssh key for authentication on linux
+	//sshPublicKeyPath := "/home/user/.ssh/id_rsa.pub"
+	//var sshBytes []byte
```
tylerharter, line 653: Delete commented out code

tylerharter (changes requested):

src/boss/cloudvm/worker.go
```diff
@@ -118,10 +123,19 @@ func (pool *WorkerPool) startNewWorker() {
...
 		worker.numTask = 1
-		pool.CreateInstance(worker) //create new instance
-
-		if pool.platform != "mock" {
+		err := pool.CreateInstance(worker) //create new instance
+		// TODO: need to handle this error, not panic (may use channel?)
+		if err != nil {
+			log.Fatalf(err.Error())
+		}
+		if pool.platform == "gcp" {
```
tylerharter, line 131: why do we need different logic for these two?  Can we handle errors consistently?

src/boss/cloudvm/worker.go
```diff
@@ -339,7 +353,12 @@ func (w *Worker) runCmd(command string) {
 
 	tries := 10
 	for tries > 0 {
-		sshcmd := exec.Command("ssh", user.Username+"@"+w.workerIp, "-o", "StrictHostKeyChecking=no", "-C", cmd)
+		var sshcmd *exec.Cmd
```
tylerharter, line 356: it feels like we should have a generic ssh-with-retry kind of function used by both platforms

src/boss/config.go
```diff
@@ -16,17 +17,19 @@ type Config struct {
...
+	Azure      cloudvm.AzureConfig `json:"azure"`
+	Gcp        cloudvm.GcpConfig   `json:"gcp"`
 }
 
 func LoadDefaults() error {
 	Conf = &Config{
 		Platform:   "mock",
 		Scaling:    "manual",
-		API_key:    "abc", // TODO: autogenerate a random key
+		API_key:    "abc", // TODO
```
tylerharter, line 28: why are you deleting the more detailed comment?

tylerharter (changes requested):

PR #189, opened 2023-10-02
Compile regex expression during initialization

PrasunGuragain:


src/worker/lambda/handlerPuller.go
```diff
@@ -26,6 +26,7 @@ type HandlerPuller struct {
 	prefix   string   // combine with name to get file path or URL
 	dirCache sync.Map // key=lambda name, value=version, directory path
 	dirMaker *common.DirMaker
+	compiled *regexp.Regexp
```
tylerharter, line 29: Better name possible to indicate what it mattches?  What about something like handlerNameRegex?

tylerharter (approved):

PR #193, opened 2023-10-05
Fix: slow copy

Charlesq666:
A minor fixing to slow copy by using external package copy.

below is the flamegraph of the new lambda pulling function using copy library vs the original one. It is not a significant boost, but the os stopped to start a process when pulllocalfile
newcopy
origincopy

Added a pprof doc of how to using pprof to generating flame graph. Thanks to Zihan

Using multi threads to boost the speed of ./ol bench init as the original version took too long to generating more than 60k files and the print statements also slow down the speed.


src/worker/lambda/handlerPuller.go
```diff
@@ -156,21 +217,36 @@ func (cp *HandlerPuller) pullLocalFile(src, lambdaName string) (rt_type common.R
...
+		err := ccopy(src, filepath.Join(targetDir, "f.py"))
 		rt_type = common.RT_PYTHON
 
-		if output, err := cmd.CombinedOutput(); err != nil {
-			return rt_type, "", fmt.Errorf("%s :: %s", err, string(output))
+		if err != nil {
+			return rt_type, "", fmt.Errorf("%s :: %s", err)
 		}
+
+//		cmd := exec.Command("cp", src, filepath.Join(targetDir, "f.py"))
```
tylerharter, line 227: let's deleted commented out code

src/worker/lambda/handlerPuller.go
```diff
@@ -33,6 +33,62 @@ type CacheEntry struct {
 	path    string // where code is extracted to a dir
 }
 
+func ccopy(src, dest string) error {
```
tylerharter, line 36: not sure what the "c" stands for -- can we give it a longer name?

src/worker/lambda/handlerPuller.go
```diff
@@ -33,6 +33,62 @@ type CacheEntry struct {
...
+	info, err := os.Stat(src)
+	if err != nil {
+		return err
+	}
+
+	if info.IsDir() {
+		return filepath.Walk(src, func(path string, info os.FileInfo, err error) error {
```
tylerharter, line 43: The docs say "Walk is less efficient than WalkDir, introduced in Go 1.16, which avoids calling os.Lstat on every visited file or directory."  I'm not clear on the difference (hope you can look closer), but any reason WalkDir doesn't work?

Also, I'm wondering if WalkDir checks for you, or if you need to do this yourself

src/worker/lambda/handlerPuller.go
```diff
@@ -33,6 +33,62 @@ type CacheEntry struct {
...
+
+			return copyFile(path, filepath.Join(dest, relPath))
+		})
+	}
+
+	return copyFile(src, dest)
+}
+
+func copyFile(src, dest string) error {
+	srcFile, err := os.Open(src)
```
tylerharter, line 65: How does this work if "src" is a symlink?  Do we copy the link itself, or do we copy the data?  Copying the data seems dangerous -- somebody could add a symlink pointing to whatever on the host machine and get it inside their lambda.  Please check and verify (and leave a comment explaining the security of it).

Some tests with symlinks are a good idea too.

tylerharter (changes requested):

src/worker/lambda/handlerPuller.go
Charlesq666: Changed to WalkDir. It does say it's more efficient in the doc. Just wasn't aware this is an option when I was writing it.

src/worker/lambda/handlerPuller.go
Charlesq666: Was thinking it to represent "customeBuiledCopy". Changed the name to "Copy".

src/worker/lambda/handlerPuller.go
Charlesq666: Deleted

src/worker/lambda/handlerPuller.go
Charlesq666: That is indeed a situation not considered at all, if the src file is a symlink, it will trace to the content and copy it and make this function prone to attack. Added a check in the new version of code to check if the src is a symlink and right now copy the symlink instead of the content (or maybe we should ignore symlink?).

tylerharter: Thanks for the security fix!  Can we do it faster, though?  I see some folks combine the check with an OpenFile call, like this: https://blog.trailofbits.com/2020/11/24/smart-and-simple-ways-to-prevent-symlink-attacks-in-go.

tylerharter (changes requested):

src/worker/lambda/handlerPuller.go
Charlesq666: Thanks! I just change the symbolic link check to O_NOFOLLOW when opening the file

tylerharter (approved):
Great work!

PR #203, opened 2023-11-08
cgroup logging with slog

ikengn:


src/worker/sandbox/cgroups/cgroup.go
```diff
@@ -20,10 +22,41 @@ type CgroupImpl struct {
...
+			logFilePath := path.Join(common.Conf.Log.Log_file_dir, "log.json")
+			f, err := os.OpenFile(logFilePath, os.O_RDWR|os.O_APPEND, 0666)
+			if err != nil {
+				panic(fmt.Errorf("Cannot open log file at %s", logFilePath))
+			}
+			logger = slog.New(slog.NewJSONHandler(f, nil))
+			//defer f.Close()
+		}
+
+		if currLevel == "INFO" && level == currLevel {
```
tylerharter, line 50: The only time levels should potentially appear as strings in the code is when we're reading from the config, everywhere else we should use the different .Info, .Warn functions, etc. OR we should use Level objects: https://pkg.go.dev/log/slog#Level

src/worker/server/server.go
```diff
@@ -215,6 +216,34 @@ func Main() (err error) {
 		return err
 	}
 
+	// cleaning old logs and setting up new logs
```
tylerharter, line 219: I'm wondering if we really need to support logging to files at all at this point?  Probably something good to do eventually, but we do already have a detach mode ("-d") that launches the worker in the background and redirects stdout to worker.out

tylerharter (changes requested):
Great first pass!

src/common/config.go
```diff
@@ -175,11 +179,14 @@ func LoadDefaults(olPath string) error {
...
-			Package: false,
-			Latency: false,
+			Format:        "default",
+			Log_file_dir:  logfileDir,
+			Cgroups:       false,
+			Memory:        false,
+			Evictor:       false,
+			Package:       false,
+			Latency:       false,
+			Cgroups_level: "INFO",
```
tylerharter, line 189: Can we just have one setting, not both Cgroups and Cgroups_level?  Perhaps Cgroups_level being set to "" can be used to indicate it is disabled.

src/common/config.go
```diff
@@ -248,6 +255,16 @@ func checkConf() error {
 		return fmt.Errorf("Unknown Sandbox type '%s'", Conf.Sandbox)
 	}
 
+	if Conf.Trace.Format != "default" {
+		if Conf.Trace.Format == "text" || Conf.Trace.Format == "json" {
```
tylerharter, line 259: So the two formats are "default" and "text"?  I haven't looked at this carefully, but I figured JSON would be one of the options.  I don't like "default" as one of the options because it is not obvious what it is.

src/common/config.go
```diff
@@ -248,6 +255,16 @@ func checkConf() error {
...
+			if Conf.Trace.Log_file_dir == "" {
```
tylerharter, line 260: If this is specified, can't we just send the log output to stdout?

src/common/config.go
```diff
@@ -248,6 +255,16 @@ func checkConf() error {
...
+				return fmt.Errorf("must specify log output directory under log_file_dir in config.json")
+			}
+		} else {
+			return fmt.Errorf("unknown log format")
```
tylerharter, line 264: Best to include the unknown name in the error message

src/common/config.go
```diff
@@ -78,11 +78,14 @@ type FeaturesConfig struct {
 }
 
 type TraceConfig struct {
-	Cgroups bool `json:"cgroups"`
-	Memory  bool `json:"memory"`
-	Evictor bool `json:"evictor"`
-	Package bool `json:"package"`
-	Latency bool `json:"latency"`
+	Format        string `json:"format"`
+	Log_file_dir  string `json:"log_file_dir"`
```
tylerharter, line 82: Do we need a separate directory for this, or just a path?  Is there more than one file?

src/worker/sandbox/cgroups/cgroup.go
```diff
@@ -43,7 +37,9 @@ func (cg *CgroupImpl) Release() {
 				if i == 0 {
 					panic(fmt.Errorf("Cannot release cgroup that contains processes: %v", pids))
 				} else {
-					cg.printf("cgroup Rmdir failed, trying again in 5ms")
+					if common.Conf.Trace.Cgroups {
```
tylerharter, line 40: do we need to check this each time?  Can't we just have a logger that somehow drops the log messages if it's not enabled?

src/worker/sandbox/cgroups/cgroup.go
```diff
@@ -53,27 +49,35 @@ func (cg *CgroupImpl) Release() {
 
 		select {
 		case cg.pool.recycled <- cg:
-			cg.printf("release and recycle")
+			if common.Conf.Trace.Cgroups {
+				cg.logger.Info("release and recycle", "CGROUP Pool", cg.pool.Name, "CGROUP", cg.name)
```
tylerharter, line 53: Can't you use a logger.With call when creating the logger so we always get this info automatically?

src/worker/sandbox/cgroups/cgroup.go
```diff
@@ -53,27 +49,35 @@ func (cg *CgroupImpl) Release() {
...
+			}
 			return
 		default:
 		}
 	}
 
-	cg.printf("release and Destroy")
+	if common.Conf.Trace.Cgroups {
+		cg.logger.Info("release and Destroy", "CGROUP Pool", cg.pool.Name, "CGROUP", cg.name)
```
tylerharter, line 61: For keys: no spaces, and no upper case

src/worker/sandbox/cgroups/pool.go
```diff
@@ -23,20 +22,52 @@ type CgroupPool struct {
 	recycled chan *CgroupImpl
 	quit     chan chan bool
 	nextID   int
+	logger   slog.Logger
 }
 
 func NewCgroupPool(name string) (*CgroupPool, error) {
+
+	level := new(slog.LevelVar)
```
tylerharter, line 30: Won't this kind of logic be needed in lots of places?  Best to do at the "common" level

src/worker/sandbox/cgroups/pool.go
```diff
@@ -23,20 +22,52 @@ type CgroupPool struct {
...
+		level.Set(slog.LevelWarn)
+	} else if common.Conf.Trace.Cgroups_level == "ERROR" {
+		level.Set(slog.LevelError)
+	}
+	// Default logger right now can't have the level be changed (haven't found the API for that, probably have to implement one from the interface)
+	logger := slog.Default()
+
+	// Replacing default logger based on config
+	if common.Conf.Trace.Format == "text" {
+		logFilePath := path.Join(common.Conf.Trace.Log_file_dir, "log.txt")
```
tylerharter, line 43: Shouldn't this be created at a much more general level?  I imagine this log file is shared across many different subsystems.

tylerharter, line 43: How about "worker.log" and don't change the extension?

src/worker/sandbox/sock.go
```diff
@@ -3,14 +3,14 @@ package sandbox
 import (
```
tylerharter, line 3: Let's just iterate on cgroups until we're happy with how we want to do logging.  Other subsystems (like SOCK) should be other pull requests.

src/worker/server/server.go
```diff
@@ -215,6 +217,36 @@ func Main() (err error) {
 		return err
 	}
 
+	// clean up old logs & set up new logs
```
tylerharter, line 220: Do we need all this complicated cleanup stuff?  I would just have a single log path, and whenever starting, open that file in an overwrite mode.

tylerharter (changes requested):

src/common/config.go
ikengn: tell them exactly what they typed

src/common/config.go
ikengn: default to the stdout, else json to a file

src/common/config.go
ikengn: reuse the worker.out

src/common/config.go
```diff
@@ -175,11 +179,14 @@ func LoadDefaults(olPath string) error {
...
 		},
 		Trace: TraceConfig{
-			Cgroups: false,
-			Memory:  false,
-			Evictor: false,
-			Package: false,
-			Latency: false,
+			Format:        "default",
+			Log_file_dir:  logfileDir,
+			Cgroups:       false,
```
tylerharter, line 184: consolidate to one setting, don't have separate Cgroups and Cgroups_level (just the latter).

src/common/logger.go
```diff
@@ -0,0 +1,44 @@
...
+	"fmt"
+	"log/slog"
+	"os"
+	"path"
+
+	"github.com/urfave/cli/v2"
+)
+
+// the current slog does not support enable/disable feature, might need a wrap-arround struct
+var CgTopLogger slog.Logger
```
tylerharter, line 13: Nothing in common should know about subsystems such as control groups.  How about just TopLogger, and subsystems can create their own loggers off of that?

src/common/logger.go
```diff
@@ -0,0 +1,44 @@
...
+
+func LoadLoggers() error {
+	if Conf.Trace.Enable_JSON == true {
+		level := new(slog.LevelVar)
+		if Conf.Trace.Cgroups == "INFO" {
```
tylerharter, line 18: Please create a level string parsing function.  It should not be case sensitive.

"" and "off" should mean it's off.  Level is just an int, so we can use LevelError+1 for off, since no logs will be high enough threshold.

src/worker/sandbox/cgroups/cgroup.go
```diff
@@ -18,13 +18,7 @@ type CgroupImpl struct {
...
 	pool       *CgroupPool
 	memLimitMB int
-}
-
-func (cg *CgroupImpl) printf(format string, args ...any) {
-	if common.Conf.Trace.Cgroups {
-		msg := fmt.Sprintf(format, args...)
-		log.Printf("%s [CGROUP %s: %s]", strings.TrimRight(msg, "\n"), cg.pool.Name, cg.name)
-	}
+	logger    	slog.Logger
```
tylerharter, line 21: We'll be typing this a lot, so lets just have "log" instead of "logger"

src/worker/sandbox/cgroups/pool.go
```diff
@@ -23,20 +22,24 @@ type CgroupPool struct {
...
+
 	pool := &CgroupPool{
 		Name:     path.Base(path.Dir(common.Conf.Worker_dir)) + "-" + name,
 		ready:    make(chan *CgroupImpl, CGROUP_RESERVE),
 		recycled: make(chan *CgroupImpl, CGROUP_RESERVE),
 		quit:     make(chan chan bool),
 		nextID:   0,
+		logger:   common.CgTopLogger,
 	}
+	pool.logger = *pool.logger.With("cgroup pool", pool.Name)
```
tylerharter, line 38: so spaces in key values.  How about "cg-pool"

src/common/config.go
ikengn: My bad, I forgot to push the latest config.json as well. This is an outdated version of the config file.

src/worker/sandbox/cgroups/pool.go
```diff
@@ -23,20 +22,24 @@ type CgroupPool struct {
...
 
 func NewCgroupPool(name string) (*CgroupPool, error) {
+
 	pool := &CgroupPool{
 		Name:     path.Base(path.Dir(common.Conf.Worker_dir)) + "-" + name,
 		ready:    make(chan *CgroupImpl, CGROUP_RESERVE),
 		recycled: make(chan *CgroupImpl, CGROUP_RESERVE),
 		quit:     make(chan chan bool),
 		nextID:   0,
+		logger:   common.CgTopLogger,
```
tylerharter, line 36: why not just set it to what you want instead of immediately updating it after creating the struct?

src/worker/sandbox/cgroups/pool.go
```diff
@@ -56,32 +59,28 @@ func (pool *CgroupPool) NewCgroup() Cgroup {
 	pool.nextID++
 
 	cg := &CgroupImpl{
-		name: fmt.Sprintf("cg-%d", pool.nextID),
-		pool: pool,
+		name:   fmt.Sprintf("cg-%d", pool.nextID),
+		pool:   pool,
+		logger: common.CgTopLogger,
 	}
+	cg.logger = *cg.logger.With("cgroup pool", cg.pool.Name, "cgroup", cg.name)
```
tylerharter, line 66: can't you create if from the pools logger so you don't need to specify the pool name again?

src/worker/sandbox/sock.go
```diff
@@ -3,14 +3,14 @@ package sandbox
...
-	"log"
+	"log/slog"
 	"sync/atomic"
 	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strconv"
-	"strings"
+	//"strings"
```
tylerharter, line 13: just delete, don't comment stuff out

src/worker/sandbox/sock.go
```diff
@@ -44,8 +44,12 @@ type SOCKContainer struct {
 // add ID to each log message so we know which logs correspond to
 // which containers
 func (container *SOCKContainer) printf(format string, args ...any) {
```
tylerharter, line 46: doesn't this function just go away?

src/common/logger.go
```diff
@@ -0,0 +1,44 @@
...
+		}
+		logFilePath := path.Join(olpath, "worker.json")
+		f, err := os.OpenFile(logFilePath, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)
+		if err != nil {
+			panic(fmt.Errorf("Cannot open log file at %s", logFilePath))
+		}
+		CgTopLogger = *slog.New(slog.NewJSONHandler(f, &slog.HandlerOptions{Level: level}))
+
+	} else {
+		// default slog's logger does not support level configuration at the moment
```
tylerharter, line 37: then let's not use it

tylerharter (changes requested):
Good progress!

ikengn:
I left a few comments in logger.go right next to the code to make it more clear. Could you take a look at them and let me know your decision for those parts, please? Also, I am currently working on the CGroup subsystem only. Other subsystems such as Sock are not being dealt with yet for now. As soon as you approve all the changes, I will go ahead and apply them to the rest of the OLs, which can then wrap up the project.

PR #204, opened 2023-12-07
Port wasm-worker to wasmtime and some other improvements

kaimast:
This ports the wasm-worker from wasmer to wasmtime. The main advantage of this change is that we do not depend on LLVM anymore (faster builds and fewer causes for build errors) and all outdated dependencies are updated (no more Dependabot warnings).

This pull request also adds two other features to the wasm-worker:
* A HTTP request API: Allows easy interaction with other services such as a database
* Setting configuration values: Allows to, for example, set the address of the database


src/worker/lambda/lambdaFunction.go
```diff
@@ -303,9 +303,9 @@ func (f *LambdaFunc) Task() {
 
 		// AUTOSCALING STEP 1: decide how many instances we want
 
-		// let's aim to have 1 sandbox per second of outstanding work
+		// let's aim to have 1 sandbox per 10ms of outstanding work
 		inProgressWorkMs := outstandingReqs * execMs.Avg
-		desiredInstances := inProgressWorkMs / 1000
+		desiredInstances := inProgressWorkMs / 10
```
kaimast, line 308: We talked about this before. I think either the policy itself should be configurable, or we should be able to set this parameter instead of hardcoding it.

kaimast:
@tylerharter can you take a look at the comments I left before I merge this?

There are a lot of commits because it is based on the branch I have been using for the last year or so, but the actual changes to the Go code are not that significant.

src/worker/server/lambdaServer.go
```diff
@@ -43,7 +43,7 @@ func (s *LambdaServer) RunLambda(w http.ResponseWriter, r *http.Request) {
 	t := common.T0("web-request")
 	defer t.T1()
 
-	log.Printf("Received request to %s\n", r.URL.Path)
+	//log.Printf("Received request to %s\n", r.URL.Path)
```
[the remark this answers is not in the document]
tylerharter: Sounds good!  @ikengn is working on moving to slog and making logging levels configurable: https://github.com/open-lambda/open-lambda/pull/203.  We can bring it back then.

src/worker/lambda/lambdaFunction.go
tylerharter: Sounds good!  Should we add a TODO that we should make it configurable at some point?

src/worker/sandbox/safeSandbox.go
```diff
@@ -118,7 +118,7 @@ func (sb *safeSandbox) DestroyIfPaused(reason string) {
 }
 
 func (sb *safeSandbox) Pause() (err error) {
-	sb.printf("Pause()")
+    //sb.printf("Pause()")
```
tylerharter, line 121: let's leave these in, because the config already lets you disable all cg logging.

src/worker/sandbox/safeSandbox.go
```diff
@@ -141,7 +141,7 @@ func (sb *safeSandbox) Pause() (err error) {
 }
 
 func (sb *safeSandbox) Unpause() (err error) {
-	sb.printf("Unpause()")
+	//sb.printf("Unpause()")
```
tylerharter, line 144: same

src/worker/sandbox/sock.go
```diff
@@ -119,10 +119,15 @@ func (container *SOCKContainer) launchContainerProxy() (err error) {
 	var procAttr os.ProcAttr
 	procAttr.Files = []*os.File{os.Stdin, os.Stdout, os.Stderr}
 
-	proc, err := os.StartProcess("./ol-container-proxy", args, &procAttr)
+    binPath, err := exec.LookPath("ol-container-proxy")
+	if err != nil {
```
tylerharter, line 123: spacing...

src/worker/lambda/lambdaManager.go
```diff
@@ -112,9 +112,9 @@ func (mgr *LambdaMgr) Get(name string) (f *LambdaFunc) {
 		f = &LambdaFunc{
 			lmgr:      mgr,
 			name:      name,
-			funcChan:  make(chan *Invocation, 32),
-			instChan:  make(chan *Invocation, 32),
-			doneChan:  make(chan *Invocation, 32),
+			funcChan:  make(chan *Invocation, 1024),
```
tylerharter, line 115: let's add a TODO to make it configurable

PR #207, opened 2023-12-30
replaced Compile() with MustCompile() and supressed panic

zeeliu7:
What to expect in the new code:
Best case scenario: regexp.MustComple() does not raise panic
Acceptable scenario: the raised panic does not create an effect on the execution
Worst case scenario: encountered a runtime error and the panic can be viewed from the terminal

(P.S. I only edited useRegex() and beginning part of Pull(). I suppose VScode stored previous edits as well. Please consider double-checking other functions if necessary.)


zeeliu7:
Added additional test to check for regex compilation inside pull(), removed panic suppression

src/worker/lambda/handlerPuller.go
```diff
@@ -106,14 +105,41 @@ func (cp *HandlerPuller) isRemote() bool {
...
+	errorMsg := error(nil)
+
+	if err != nil {
+		msg := "regex failed to comple with error '%s'"
+		errorMsg = fmt.Errorf(msg, err)
+	}
+	if re == nil {
+		errorMsg = errors.New("regex failed to compile with null output")
+	} else {
+		vMatched := re.MatchString(testValidString)
```
tylerharter, line 119: I don't think we want to do testing every time (will slow things down)

src/worker/lambda/handlerPuller.go
```diff
@@ -106,14 +105,41 @@ func (cp *HandlerPuller) isRemote() bool {
...
+
 func (cp *HandlerPuller) Pull(name string) (rt_type common.RuntimeType, targetDir string, err error) {
 	t := common.T0("pull-lambda")
 	defer t.T1()
 
-	matched, err := regexp.MatchString(`^[A-Za-z0-9\.\-\_]+$`, name)
-	if err != nil {
-		return rt_type, "", err
-	} else if !matched {
+	handlerNameRegex, handlerNameError := createRegex(`^[A-Za-z0-9\.\-\_]+$`, "-H3ll0.world_", "Hello world!")
```
tylerharter, line 136: this will compile a new regex every time, so we won't have any performance benefit.  We want to compile once and have it in a global variable or somewhere so that we can keep reusing with recompiling.

tylerharter (changes requested):

PR #208, opened 2024-01-03
Enable basic go linting

kaimast:
This should make reviewing pull request a little easier.

For now, most lints are still disabled. Enabling them and fixing lint failures on the existing code base might be a good first issue for students.


tylerharter:
Cool, thanks for adding this!

PR #209, opened 2024-01-03
Added one-term RegEx test and evaluation for Pull() usage

zeeliu7:
This might be the only way I come up to ensure it only compiles and tests once, but I am willing to discuss about other methods.


src/worker/lambda/handlerPuller.go
```diff
@@ -21,6 +21,11 @@ import (
 
 var notFound404 = errors.New("file does not exist")
 
+// for regex inside Pull() use
+var handlerNameRegex *regexp.Regexp
```
tylerharter, line 25: Instead of the 3 variables, can we just have this?

var handlerNameRegex = regexp.MustCompile(`^[A-Za-z0-9\.\-\_]+$`)

src/worker/lambda/handlerPuller.go
```diff
@@ -106,14 +111,46 @@ func (cp *HandlerPuller) isRemote() bool {
...
+	if handlerNameRegexStatus == 0 {
+		createRegex(`^[A-Za-z0-9\.\-\_]+$`, `H3ll0.world_`, `Hello world!`)
+	}
+
+	// return error if failed to compile
+	if handlerNameRegexStatus == -1 {
+		return rt_type, "", handlerNameRegexError
+	}
+
+	matched := handlerNameRegex.MatchString(name)
```
tylerharter, line 152: this is probably short enough to use in the next line, with the if (no "matched" variable needed)

tylerharter (changes requested):

zeeliu7:
GitHub says there is conflict with the file. I added one local variable "var handlerNameRegex *regexp.Regexp". Please let me know if there are problems.

PR #210, opened 2024-01-03
Update handlerPuller.go with compiled regex

zeeliu7:
now using a compiled regex with panic for errors

### DIFF
--- a/src/worker/lambda/handlerPuller.go
+++ b/src/worker/lambda/handlerPuller.go
@@ -20,7 +20,7 @@ import (
 )
 
 var errNotFound404 = errors.New("file does not exist")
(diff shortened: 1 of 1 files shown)


src/worker/lambda/handlerPuller.go
```diff
@@ -20,6 +20,7 @@ import (
 )
 
 var errNotFound404 = errors.New("file does not exist")
+var handlerNameRegex *regexp.Regexp
```
tylerharter, line 23: let's do MustCompile up here instead

tylerharter (changes requested):

zeeliu7:
Fixed compile-once problem with local variable

tylerharter (approved):

PR #213, opened 2024-01-03
Wsgi Draft

liamjogal:
By modifying the server.py code to include a child class from tornado.wsgi.WSGIContainer, I modified a method in the WSGIContainer class. This changes routing, so that clients that request for pages in a flask app like "run/lambda/helloworld" would actually be requesting for the page "helloworld" in the flask app. Also, this wouldn't work for apps that had routes outside of "/" so I modified code in lambdaServer.go.


src/worker/server/lambdaServer.go
```diff
@@ -60,8 +60,11 @@ func (s *LambdaServer) RunLambda(w http.ResponseWriter, r *http.Request) {
 			img := urlParts[1]
 			s.lambdaMgr.Get(img).Invoke(w, r)
 		} else {
-			w.WriteHeader(http.StatusInternalServerError)
-			w.Write([]byte("expected invocation format: /run/<lambda-name>"))
+			img := urlParts[1]
+			s.lambdaMgr.Get(img).Invoke(w, r)
```
tylerharter, line 64: If the two branches of the if/else are the same, can we simplify down to one code path?

Hopefully we can retain the deleted error message, but do it in server.py when it's not WSGI and we don't have the expected number of parts in the URL.

PR #215, opened 2024-01-04
update go version in directions since we need at least 1.20

tylerharter:


kaimast:
Thanks! I noticed that I accidentally changed the minimum go version after updating the docker dependency....

PR #216, opened 2024-01-05
add trace playing capability

tylerharter:


kwangjong:
Everything looks great to me! The only thing I added is filtering out empty lines because I encountered an error when the trace file contained a trailing newline.

PR #218, opened 2024-01-09
Evicting packages

pt-07:


src/worker/lambda/packages/packagePuller.go
```diff
@@ -8,7 +8,10 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"os/exec"
 	"path/filepath"
+	//test
```
tylerharter, line 13: cleanup

src/worker/lambda/packages/packagePuller.go
```diff
@@ -21,6 +24,48 @@ import (
...
+var sizeTracker = SizeOfPackages{
+	TotalSize: 0,
+	Packages:  make(map[string]int),
+}
+
+func PackageSizeIncreaser(pkg string, size int) {
+	// Normalize the package name
+	pkg = NormalizePkg(pkg)
+
+	// Lock the sizeTracker here if using in a concurrent context
```
tylerharter, line 43: need actual lock

src/worker/lambda/packages/packagePuller.go
```diff
@@ -21,6 +24,48 @@ import (
...
+
+	// Update the size for the individual package and the total size
+	sizeTracker.Packages[pkg] += size
+	sizeTracker.TotalSize += size
+
+	// Unlock the sizeTracker here if using in a concurrent context
```
tylerharter, line 49: better to use defer style for unlocking

src/worker/lambda/packages/packagePuller.go
```diff
@@ -21,6 +24,48 @@ import (
...
+}
+
+func getPackageSize(packageName string) (string, error) {
+	cmd := exec.Command("apt-cache", "show", packageName)
```
tylerharter, line 53: does this work for PyPI packages?  I would have expected some kind of recursive directory size code or something like that...

src/worker/lambda/packages/packagePuller.go
```diff
@@ -21,6 +24,48 @@ import (
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err := cmd.Run()
```
tylerharter, line 56: do we need to launch a separate program to check size?  That's been shown as an excessive overhead in other cases...

src/worker/lambda/packages/packagePuller.go
```diff
@@ -126,6 +171,7 @@ func (pp *PackagePuller) InstallRecursive(installs []string) ([]string, error) {
 // will never try more after the first success
 func (pp *PackagePuller) GetPkg(pkg string) (*Package, error) {
 	// get (or create) package
+	pkgname = pkg
```
tylerharter, line 174: does this run?  Do we need it?

src/worker/lambda/packages/packagePuller.go
```diff
@@ -148,6 +194,7 @@ func (pp *PackagePuller) GetPkg(pkg string) (*Package, error) {
 		return p, nil
 	}
 
+	getPackageSize(pkgname)
```
tylerharter, line 197: Should you use the returned value for something?

tylerharter (changes requested):
Good progress!

PR #225, opened 2024-03-06
Package Eviction

pt-07:
Added code to correctly output package sizes


src/worker/lambda/packages/packagePuller.go
```diff
@@ -175,6 +179,29 @@ func (pp *PackagePuller) sandboxInstall(p *Package) (err error) {
 			return err
 		}
 	}
+	//gets the size of the current package
+	cmd := exec.Command("du", "-sb", scratchDir)
```
tylerharter, line 183: let's compute the size without running an external program

src/worker/lambda/packages/packagePuller.go
```diff
@@ -175,6 +179,29 @@ func (pp *PackagePuller) sandboxInstall(p *Package) (err error) {
...
+	var out bytes.Buffer
+	cmd.Stdout = &out
+	err = cmd.Run()
+	if err != nil {
+		log.Fatal(err)
```
tylerharter, line 188: Fatal will kill the server.  I don't think that's a proportionate response to an error.  Better to return it back.

src/worker/lambda/packages/packagePuller.go
```diff
@@ -12,7 +12,9 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
-
+	"regexp"
+    "strconv"
```
tylerharter, line 16: spacing

src/worker/lambda/packages/packagePuller.go
```diff
@@ -36,8 +38,10 @@ type Package struct {
 	Meta         PackageMeta
 	installMutex sync.Mutex
 	installed    uint32
+	Size int
```
tylerharter, line 41: Will need to be protected by some lock.  Perhaps with installMutex?  Which would then need to be renamed to reflect the broader usage

src/worker/lambda/packages/packagePuller.go
```diff
@@ -36,8 +38,10 @@ type Package struct {
...
 }
 
+var totalPackageSize = 0
```
tylerharter, line 44: If we have a total size variable like this, it will need to be lock protected.  Perhaps we can keep this PR simple, and figure out those details in a future PR that actually implements eviction?

src/worker/lambda/packages/packagePuller.go
```diff
@@ -175,6 +179,29 @@ func (pp *PackagePuller) sandboxInstall(p *Package) (err error) {
 			return err
 		}
 	}
+	//gets the size of the current package
```
tylerharter, line 182: let's have a config option so we don't compute/log size unless doing so is enabled

tylerharter (changes requested):
Great progress!

PR #231, opened 2024-04-23
renaming from server to event

zeeliu7:
Renaming from server to event, including all files inside src/worker/event and src/common/config.go and scripts/sock_test.py.
Kafka code coming later this week.

### DIFF
--- a/scripts/sock_test.py
+++ b/scripts/sock_test.py
@@ -56,7 +56,7 @@ def sock_churn(baseline, procs, seconds, fork):
 def run_tests():
     print("Testing SOCK directly (without lambdas)")
 
-    with TestConfContext(server_mode="sock", mem_pool_mb=500):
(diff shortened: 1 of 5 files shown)


tylerharter:
I don't think we want to rename server to event in so many places in the code (a lot of the things you renamed are genuinely servers).  I was just imagining you would rename the directory, and the go package name.  Servers are one of the things that should be in the event package.

There's some commented out code here related to Kafka too -- that shouldn't be part of it.

zeeliu7:
Dear Professor Harter,

Thanks for the reply and sorry about the mistake. In my next update I will do the renaming and import, as well as putting Kafka code at the right place. Sorry about the lack of communication over previous weeks: I was busy with finals and then my internship started immediately. I will email you when an update is available.

Thank you for your help.

Yours sincerely,
Zhonghao Liu
________________________________

PR #238, opened 2024-05-08
Fix the missing slash in the file path concatenation

vcodx:
This adds a missing slash in the file path concatenation in HandlerPuller, fixes pulling remote files named _*.tar.gz_.

In this instance, `targetDir` does not have a trailing slash, and when concatenated with `f.py`, the resulting path does never exist.


tylerharter:
Thanks for catching that and for the fix!!!

PR #240, opened 2024-05-27
Create basic automatic cleanup

nannoda:


src/worker/commands.go
```diff
@@ -229,60 +229,9 @@ func cleanupCmd(ctx *cli.Context) error {
...
-		}
-	}
-
-	if err := syscall.Unmount(dirName, syscall.MNT_DETACH); err != nil {
-		fmt.Printf("could not unmount %s: %s\n", dirName, err.Error())
-	}
-
-	if err := os.Remove(filepath.Join(olPath, "worker", "worker.pid")); err != nil {
-		fmt.Printf("could not remove worker.pid: %s\n", err.Error())
+	if err := cleanupOL(olPath); err != nil {
```
tylerharter, line 232: let's just return cleanupOL(olPath) instead of having the "if", unless the logic becomes more complicated later

src/worker/helpers.go
```diff
@@ -170,7 +170,7 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
 // main error scenarios:
 // 1. PID exists, but process cannot be killed (worker probably died unexpectedly)
 // 2. The cleanup is taking too long (maybe the timeout is insufficient, or there is a deadlock)
-func stopOL(_ string) error {
+func stopOL(olPath string) error {
```
tylerharter, line 173: let's add comments below indicating each of the scenarios we're handling:
1. worker already stopped
2. clean shutdown (via kill signal)
3. dirty shutdown (there's a worker.pid, but no corresponding process)

src/worker/helpers.go
```diff
@@ -193,7 +193,12 @@ func stopOL(_ string) error {
 	}
 	fmt.Printf("Send SIGINT and wait for worker to exit cleanly.\n")
 	if err := p.Signal(syscall.SIGINT); err != nil {
-		return fmt.Errorf("Failed to send SIGINT to PID %d (%s).  May require manual cleanup.\n", pid, err.Error())
+		// Failed to send SIGINT to PID  (os: process already finished).  May require manual cleanup will happen when system force shuts down.
+		if err.Error() == "os: process already finished" {
```
tylerharter, line 197: do we need another case here to detect a dirty shutdown?  Seems like a few lines earlier we can already identify the case that worker.pid exists but os.FindProcess can't find it.  Seems like we ought to handle the dirty shutdown there instead of returning an error.

In general, I'm not sure whether string representations of errors are guaranteed to never change, so I wouldn't want to depend on that.

src/worker/helpers.go
```diff
@@ -193,7 +193,12 @@ func stopOL(_ string) error {
...
+			fmt.Printf("Unclean exit detected, trying automatic cleanup...")
+			return cleanupOL(olPath)
+		}
+		// return fmt.Errorf("Failed to send SIGINT to PID %d (%s).  May require manual cleanup.\n", pid, err.Error())
 	}
 
 	for i := 0; i < 600; i++ {
```
tylerharter, line 204: perhaps sending signal 0 is a hacky way to see if the process has exited.  I wonder if there's any reason not to use FindProcess again?  Seems that would be cleaner.

tylerharter: Perhaps before printing "OL worker process stopped successfully.\n" we should check that worker.pid was actually deleted.  Because if it's there, then the OL process might have crashed when attempting to stop.

src/worker/helpers.go
```diff
@@ -208,6 +213,63 @@ func stopOL(_ string) error {
...
+				fmt.Printf("could not remove mount dir: %s\n", err.Error())
+			}
+		}
+	}
+
+	if err := syscall.Unmount(dirName, syscall.MNT_DETACH); err != nil {
+		fmt.Printf("could not unmount %s: %s\n", dirName, err.Error())
+	}
+
+	if err := os.Remove(filepath.Join(olPath, "worker", "worker.pid")); err != nil {
```
tylerharter, line 266: I think removal of worker.pid indicates everything has been cleaned up previously.  That might not be true here because we ignore some errors above.

src/worker/helpers.go
```diff
@@ -208,6 +213,63 @@ func stopOL(_ string) error {
...
+		return fmt.Errorf("could not remove worker.pid: %s\n", err.Error())
+	}
+
+	return nil
```
tylerharter, line 270: there are a bunch of cases earlier where we log an error, but keep going.  Keep going seems like the right decision (maybe we can cleanup some things even if not everything), but we should still return an error if there was any issue earlier.  Some options: return first error, return last error, or return a new error saying something like ("there were N errors during force cleanup")

src/worker/helpers.go
```diff
@@ -208,6 +213,63 @@ func stopOL(_ string) error {
 	return fmt.Errorf("worker didn't stop after 60s")
 }
 
+func cleanupOL(olPath string) error {
```
tylerharter, line 216: Let's name this something like dirtyShutdownCleanup so it's clear we wouldn't normally do this.

tylerharter (changes requested):
The refactor looks great!  Most of my comments here are related to issues in the code prior to your refactor, but I'm hoping you can still fix some of these things up as part of this PR.  Thanks!

src/worker/helpers.go
```diff
@@ -166,46 +166,192 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+	case DirtyShutdown:
+		dirtyCleanup(olPath)
+	case Unknown:
+		return fmt.Errorf("unknown state detected")
+	}
+
+	return nil
+}
+
+// stopOL attempts to gracefully stop Open Lambda (OL) and handles errors.
```
tylerharter, line 337: These functions seem like pretty simple wrappers.  Can we get rid of them and call generalCleanup or gracefulCleanup directly?

src/worker/helpers.go
```diff
@@ -166,46 +166,192 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+				fmt.Printf("Attempting to remove %s\n", cg)
+				if err := syscall.Rmdir(cg); err != nil {
+					// Return an error if removing a cgroup fails.
+					return fmt.Errorf("could not remove cgroup: %s", err.Error())
+				}
+			}
+		}
+		if err := syscall.Rmdir(cgRoot); err != nil {
+			// Log an error if removing the cgroup root directory fails.
+			fmt.Printf("Could not remove cgroup root: %s\n", err.Error())
```
tylerharter, line 270: If this fails, we'll never return an error

src/worker/helpers.go
```diff
@@ -166,46 +166,192 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+	if files, err := os.ReadDir(dirName); err != nil {
+		// Return an error if the mount root directory cannot be found.
+		return fmt.Errorf("could not find mount root: %s", err.Error())
+	} else {
+		for _, file := range files {
+			path := filepath.Join(dirName, file.Name())
+			fmt.Printf("Attempting to unmount %s\n", path)
+			if err := syscall.Unmount(path, syscall.MNT_DETACH); err != nil {
+				// Return an error if unmounting fails.
+				return fmt.Errorf("could not unmount: %s", err.Error())
```
tylerharter, line 287: If we return now, we won't attempt to unmount the other directories (unmounting those might succeed!)

src/worker/helpers.go
```diff
@@ -166,46 +166,192 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+	Unknown = -1
+)
+
+// Check the current state of Open Lambda
+//
+// This function returns the current state of Open Lambda, the PID if possible,
+// and an error if it encounters any.
+func checkState(olPath string) (OlState, int, error) {
+	// Locate the worker.pid file, use it to get the worker's PID
 	pidPath := filepath.Join(common.Conf.Worker_dir, "worker.pid")
```
tylerharter, line 186: either use Worker_dir or olPath (probably olPath), but not both

src/worker/helpers.go
```diff
@@ -166,46 +166,192 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
-// main error scenarios:
-// 1. PID exists, but process cannot be killed (worker probably died unexpectedly)
-// 2. The cleanup is taking too long (maybe the timeout is insufficient, or there is a deadlock)
-func stopOL(_ string) error {
-	// locate worker.pid, use it to get worker's PID
+// Define a custom type for the state
+type OlState int
+
+// Define constants for the different states
+const (
```
tylerharter, line 173: Should there be an Initialized state?  Perhaps the directory hasn't even been created yet...

src/worker/helpers.go
```diff
@@ -166,46 +166,192 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+	Stopped OlState = iota
+	Running
+	DirtyShutdown
```
tylerharter, line 176: Perhaps the relationship between Stopped and DirtryShutdown could be clearer.  Maybe these names?
StoppedClean
StoppedDirty

tylerharter (changes requested):

src/worker/helpers.go
nannoda: My assumption is the user can run simply the cleanup process again after they resolve what causes those cgroups too unable to be unmounted (either though CLI or through the automatic cleanup process), and early return might be helpful for the users to find what went wrong.
I assume we don't want to simply continue when some cleanup processes were not successful.

src/worker/helpers.go
nannoda: stopOL is used in multiple different places. I think the name of generalCleanup might be a bit misleading because it also stops the process if it is running.

PR #247, opened 2024-07-08
Automatic cleanup after dirty shutdown

nannoda:


src/worker/helpers.go
```diff
@@ -166,46 +166,224 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+	fmt.Println("Attempting to gracefully shut down the worker process by sending SIGINT.")
+
+	pid, err := getPid()
 	if err != nil {
-		return fmt.Errorf("Failed to find worker process with PID %d.  May require manual cleanup.\n", pid)
+		fmt.Errorf("failed to get pid: %s", err)
 	}
-	fmt.Printf("Send SIGINT and wait for worker to exit cleanly.\n")
+
+	p, _ := os.FindProcess(pid)
```
tylerharter, line 247: My understanding is that this should never return an error.  Let's write a little snippet to panic if we do get an error for some reason.

tylerharter (changes requested):
Fantastic work!  This is very close.

src/worker/helpers.go
```diff
@@ -166,46 +166,224 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+// Define a custom type for the state
+type OlState int
+
+// Define constants for the different states
+const (
+	Uninitialized OlState = iota
+	Running
+	StoppedClean
+	StoppedDirty
+	Unknown = -1
```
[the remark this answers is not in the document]
nannoda: What should I return in check state if there is an error though? I cannot simply return nil because OlState is a number.
Would you suggest me to simply return -1?

src/worker/helpers.go
```diff
@@ -166,46 +166,224 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+			fmt.Printf("%d error(s) while cleaning up sandboxes.\n", sandboxErrorCount)
+		}
+		fmt.Printf("You can try to rerun the cleanup process again later.\n")
+		return fmt.Errorf("%d error(s) while cleaning up cgroup and %d error(s) while cleaning up sandbox", cgroupErrorCount, sandboxErrorCount)
+	}
+
+	// Attempt to unmount the main mount directory
+	if err := syscall.Unmount(dirName, syscall.MNT_DETACH); err != nil {
+		// Log an error if unmounting the main directory fails.
+		fmt.Printf("Could not unmount %s: %s\n", dirName, err.Error())
```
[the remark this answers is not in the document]
nannoda: Because when system shuts down while OL is running all those mounts will be unmounted, and calling that will result in an error. But I do agree this might not be an optimal implementation as we might encounter other errors while unmounting.

src/worker/helpers.go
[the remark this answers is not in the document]
tylerharter: Good point!

src/worker/helpers.go
```diff
@@ -166,46 +169,262 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
-// main error scenarios:
-// 1. PID exists, but process cannot be killed (worker probably died unexpectedly)
-// 2. The cleanup is taking too long (maybe the timeout is insufficient, or there is a deadlock)
-func stopOL(_ string) error {
-	// locate worker.pid, use it to get worker's PID
+// Define a custom type for the state
+type OlState int
+
+// Define constants for the different states
+const (
```
tylerharter, line 176: We should have a comment briefly describing each of the states

src/worker/helpers.go
```diff
@@ -166,46 +169,262 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+	Uninitialized OlState = iota
+	Running
+	StoppedClean
+	StoppedDirty
+	Unknown = -1
+)
+
+// Check the current state of OpenLambda
+//
+// This function returns the current state of OpenLambda, the PID if possible,
```
tylerharter, line 186: It doesn't return a PID though, does it?

src/worker/helpers.go
```diff
@@ -166,46 +169,262 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+		fmt.Println("Successfully cleaned up from the dirty shutdown.")
+	case StoppedClean:
+		fmt.Println("No OpenLambda instance is running. No further actions are needed.")
+	case Uninitialized:
+		fmt.Println("OpenLambda is not initialized. You should initialized it.")
+		return fmt.Errorf("cannot bring Uninitialized to StoppedClean")
+	default:
+		return fmt.Errorf("unrecognized state")
+	}
+
```
tylerharter, line 426: A good cautious thing to do would be to call checkState again and make sure we actually are in StoppedClean now.  We used to have a lot of bugs where a clean shutdown wouldn't actually cleanup everything, so I'm always thinking about how we'll just similar bugs if they arise in the future.

src/worker/helpers.go
```diff
@@ -166,46 +169,262 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
...
+	if err := os.Remove(filepath.Join(olPath, "worker", "worker.pid")); err != nil {
+		// Return an error if removing worker.pid fails.
+		return fmt.Errorf("could not remove worker.pid: %s", err.Error())
+	}
+
+	return nil
+}
+
+// bringToStoppedClean tries the best to bring the state of OpenLambda to StoppedClean no mater which state it is in.
+func bringToStoppedClean(olPath string) error {
```
tylerharter, line 397: should we have something like bringToRunning too?

src/worker/commands.go
```diff
@@ -229,61 +229,7 @@ func cleanupCmd(ctx *cli.Context) error {
...
-	if err := syscall.Unmount(dirName, syscall.MNT_DETACH); err != nil {
-		fmt.Printf("could not unmount %s: %s\n", dirName, err.Error())
-	}
-
-	if err := os.Remove(filepath.Join(olPath, "worker", "worker.pid")); err != nil {
-		fmt.Printf("could not remove worker.pid: %s\n", err.Error())
-	}
-
-	return nil
+	return bringToStoppedClean(olPath)
```
tylerharter, line 232: is cleanupCmd any different than downCmd?  I guess downCmd loads the configs, but we don't actually use it, right?

What would you suggest in the end?  One command, or both?  And what would the semantics of them be?

tylerharter (approved):
Great work!  I left some comments here, but perhaps you could work on them in another PR (if you're interested in doing so).  I think we should merge this one in.  It's already a huge improvement in how we handle state and cleanup.

PR #251, opened 2024-08-21
add debug tutorial

ShockYoungCHN:
Hi Tyler, please check out the debug tutorial, I deal with trademarks carefully.


tylerharter:
Thanks!  I'm wondering if the OL repo is the right place for this, though?  It seems to provide background on Go debug/tooling in general.  I could see some docs specifically related to OL debugging, but I'm reluctant to add docs about more general tooling (because then we'll need to update those when it inevitably becomes stale, and staying on top of OL-specific documentation will be effort enough).

PR #258, opened 2024-11-09
Fix `use cgroup` log message

juzeon:
This PR fixes a minor bug of a log message.

Before:

```
use cgroup %!s(func() string=0x8e1860) [SOCK 1]
```

After:

```
use cgroup cg-1 [SOCK 1]
```


kaimast:
Looks good. Thanks for the pull request!

PR #261, opened 2024-12-08
Working mirror tested only with filesystem mirror

Siggyv:
Edited packagePuller.go, handlerPuller.go and packagePullerInstaller.py

packagePuller.go:
Added pip_mirror configuration field to message sent to sandbox. Allows packagePullerInstaller.py to see if there is a pip mirror installed or not, and if so where.

handlerPuller.go:
Was not originally pulling the requirements.txt file from the registry to the targetDir. This caused any requested downloads with requirements.txt to be ignored and not ran with or without a pip mirror indicated as set up.

packagePullerInstaller.py:
Changed so that if a destination for a backup mirror was specified, pip then installs only from the mirror with --find-links. --find-links may be able to work for docker or http servers. That is the next step.

TL;DR
Works for filesystem based pip mirrors. However, with some additional testing, it may also work for docker or http servers.

### DIFF
--- a/docs/worker/pip-mirror.md
+++ b/docs/worker/pip-mirror.md
@@ -1,0 +1,100 @@
+# Installing a Pip Mirror
+
+When running python based lambdas in OpenLambda packages
+are installed with pip. By default, pip installs packages
+through an API to pypi.org.
+
+Pypi.org cannot guarantee that every indexed package is safe,
+thus a local mirrored index for pip may be beneficial.
+
+### Creating and Defining the Bandersnatch Mirror
+
+In order to setup a bandersnatch pip mirror first install bandersnatch with:
+```
+pip install bandersnatch
+```
+
+Once you have installled bandersnatch, you will have to create the config file.
+To do this run:
+```
+bandersnatch mirror
+```
+
+This will create a bandersnatch.conf file. In this configuration
+file you can set different settings for the mirror.
+
+Furthermore, you can decide the backend for the bandersnatch mirror.
+This can be a docker container, web server, or a filesystem.
+
+For documentation on how to set up the mirror use: [bandersnatch docs](https://bandersnatch.readthedocs.io/en/latest/mirror_configuration.html)
+
+It is important to note that if you do not use an allowlist or 
+blacklist for configuration you will be installing the whole 
+pypi index which is several terabytes.
+
+When deciding which packages to install into the mirror it is important
+to note that the current implementation of the package puller, when
+a pip mirror is defined, is to not use pypi.org even upon install
+failure.
+
+Furthermore, since OpenLambda does not install dependencies with 
+pip's `--no-deps` please ensure that any packages you wish to 
+install have all required dependencies
+within the mirror as well. For more information on how OpenLambda
+installs packages view: [pypi-packages](pypi-packages.md). 
+
+Once you have set up the bandersnatch mirror and are happy with the
+configuration file you can run `bandersnatch mirror` once more to 
+begin installation. This may take several minutes.
+
+### Worker Setup
+
+After the mirror has finished installation, you should initialize
+a worker with:
+```
+ol worker init -p worker_name -i base_image
+```
+You can choose the worker name and the base image to initialize.
+
+Then in `config.json` edit the `pip_mirror` field with the url or
+path to your pip mirror. Do not include simple at the end, and 
+ensure that the path is the path to base directory i.e.
+`/mirror/web`.
+
+Before starting the worker ensure that in the registry directory
+you have created a `requirements.in` file with the required packages.
+Then get the required dependencies with:
+```
+pip-compile /path/to/worker/registry/requirements.in
+```
+
+If you do not yet have `pip-compile` run:
+```
+pip install pip-tools
+```
+
+Then you can start the worker with:
+```
+ol worker up
+```
+
+This will not start any installs until you have sent it a 
+POST request. 
+
+### Potential Issues
+
+If you run into an error it will likely present as unable to find
+/host/files. This is because the install failed leading to the target
+directory for packages to be undefined. 
+
+Thus some possible causes may be:
+1. Package requested is not in the mirror.
+2. Incorrect pip mirror url or directory. This may be due to read access for the mirror or web directory.
+3. Packages not including version numbers.
+4. Updates in requirements for a package causing the mirror to become outdated.
+
+To better analyze this add `-vvv` to the pip install in packagePuller.py in 
+`open-lambda/src/worker/embedded/`.
+
+Additionally, installs do not yet account for platform specific dependencies or
+conditional dependencies in general
--- a/src/worker/embedded/packagePullerInstaller.py
+++ b/src/worker/embedded/packagePullerInstaller.py
@@ -62,17 +62,28 @@ def deps(dirname):
             continue
     return list(rv)
 
-
 def f(event):
+    packages_path = '/packages/'
+    if packages_path not in sys.path:
+        sys.path.append(packages_path)
+
     pkg = event["pkg"]
     alreadyInstalled = event["alreadyInstalled"]
+    pip_mirror = event.get("pip_mirror", "")
+    pip_mirror = pip_mirror.rstrip('/') + '/simple/' # make sure it ends with / and has simple at the end
     if not alreadyInstalled:
         try:
-            subprocess.check_output(
-                ['pip3', 'install', '--no-deps', pkg, '--cache-dir', '/tmp/.cache', '-t', '/host/files'])
+            if pip_mirror == "":
+                subprocess.check_output(
+                    ['pip3', 'install', '--no-deps', pkg, '--cache-dir', '/tmp/.cache', '-t', '/host/files'])
+            else:
+                pip_mirror = pip_mirror + pkg.split("==")[0]
+                cmds = ['pip3', 'install', '--no-deps', pkg, '--cache-dir', '/tmp/.cache', '-t', '/host/files', "--no-index", f"--find-links={pip_mirror}", "-vvv"]
+                print(f"[packaagePullerInstaller.py] attempting install with command: {cmds}")
+                out = subprocess.check_output(cmds)
         except subprocess.CalledProcessError as e:
-            print(f'pip install failed with error code {e.returncode}')
-            print(f'Output: {e.output}')
+            print(f'[packagePullerInstaller.py] pip install failed with error code {e.returncode}')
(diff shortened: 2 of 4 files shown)


src/worker/lambda/handlerPuller.go
tylerharter, line 1: What are you trying to accomplish with the changes to handlerPuller.go?  Do we need to do anything here if we only support mirrors over HTTP?

src/worker/embedded/packagePullerInstaller.py
```diff
@@ -62,17 +62,27 @@ def deps(dirname):
             continue
     return list(rv)
 
-
 def f(event):
     pkg = event["pkg"]
     alreadyInstalled = event["alreadyInstalled"]
+    pip_mirror = event.get("pip_mirror", "")
```
tylerharter, line 68: If we only call this from one place and we also pass it, do we need to make it optional?

src/worker/embedded/packagePullerInstaller.py
```diff
@@ -62,17 +62,27 @@ def deps(dirname):
...
     if not alreadyInstalled:
         try:
```
tylerharter, line 70: There's a lot of code in the try, but it only catches exceptions from check_output.  Let's have a single check_output in the try block.  For that matter, let's have a single check_output in general.  We can pass in a cmd variable (or similar) that gets configured differently.

src/worker/embedded/packagePullerInstaller.py
```diff
@@ -62,17 +62,27 @@ def deps(dirname):
...
-                ['pip3', 'install', '--no-deps', pkg, '--cache-dir', '/tmp/.cache', '-t', '/host/files'])
+            if pip_mirror == "":
+                subprocess.check_output(
+                    ['pip3', 'install', '--no-deps', pkg, '--cache-dir', '/tmp/.cache', '-t', '/host/files'])
+            else:
+                pip_mirror = pip_mirror.rstrip('/') + '/simple/' # make sure it ends with / and has simple at the end
+                host_start_index = pip_mirror.find('://') + 3
+                host_end_index = pip_mirror.find('/', host_start_index)
+                mirror_host = pip_mirror[host_start_index:host_end_index]
+                cmds = ['pip3', 'install', '--no-deps', pkg, '--cache-dir', '/tmp/.cache', '-t', '/host/files', f"--trusted-host={mirror_host}", f"--index-url={pip_mirror}",  "-vvv"]
```
tylerharter, line 79: We should avoid copying the original args then adding.  If we need to make future changes, that will have to happen in two places.  Better to have the basic args, then if pip_mirror is set, append the additional ones.

src/worker/embedded/packagePullerInstaller.py
```diff
@@ -62,17 +62,27 @@ def deps(dirname):
...
+                print(f"[packaagePullerInstaller.py] attempting install with command: {cmds}")
```
tylerharter, line 80: typo: packaage

src/worker/embedded/packagePullerInstaller.py
```diff
@@ -62,17 +62,27 @@ def deps(dirname):
...
-
 def f(event):
     pkg = event["pkg"]
     alreadyInstalled = event["alreadyInstalled"]
+    pip_mirror = event.get("pip_mirror", "")
     if not alreadyInstalled:
         try:
-            subprocess.check_output(
-                ['pip3', 'install', '--no-deps', pkg, '--cache-dir', '/tmp/.cache', '-t', '/host/files'])
+            if pip_mirror == "":
```
tylerharter, line 71: Just this:
if pip_mirror:
    ....

src/worker/embedded/packagePullerInstaller.py
```diff
@@ -62,17 +62,27 @@ def deps(dirname):
...
+                subprocess.check_output(
+                    ['pip3', 'install', '--no-deps', pkg, '--cache-dir', '/tmp/.cache', '-t', '/host/files'])
+            else:
+                pip_mirror = pip_mirror.rstrip('/') + '/simple/' # make sure it ends with / and has simple at the end
+                host_start_index = pip_mirror.find('://') + 3
```
tylerharter, line 76: Why do all this preprocessing?  If you just want a host name, why not require them to put a host name in the config?

docs/worker/pip-mirror.md
```diff
@@ -0,0 +1,176 @@
...
+/host/files. This is because the install failed leading to the target
+directory for packages to be undefined. 
+
+Thus some possible causes may be:
+1. Package requested is not in the mirror.
+2. Incorrect pip mirror url or directory. This may be due to read access for the mirror or web directory.
+3. Packages not including version numbers.
+4. Updates in requirements for a package causing the mirror to become outdated.
+
+To better analyze this the pip install command has `-vvv` in order to help debugging.
```
tylerharter, line 173: How would this get surfaced to the user, though?  Is it only something you would want with a mirror but not a regular install?

docs/worker/pip-mirror.md
```diff
@@ -0,0 +1,176 @@
...
+
+If you get an error 403, this is likely due to read permissions for your
+bandersnatch mirror. As nginx requires read and execute access, additionally
+nginx may want the user `www-data` to have ownership. To do this use:
+```
+sudo chmod -R 755 {path to mirror} # for read permissions
+sudo chown -R www-data:www-data {path to mirror} # for ownership
+```
+
+### Worker Setup
```
tylerharter, line 126: It feels like there is some redundancy with these:
https://github.com/open-lambda/open-lambda/blob/main/docs/worker/getting-started.md
https://github.com/open-lambda/open-lambda/blob/main/docs/worker/pypi-packages.md

Can you point to other pages (as needed) and simplify this?

docs/worker/pip-mirror.md
```diff
@@ -0,0 +1,176 @@
...
+ol worker up
+```
+
+This will not start any installs until you have sent it a 
+POST request. 
+
+### Potential Issues
+
+If you run into an error it will likely present as unable to find
+/host/files. This is because the install failed leading to the target
```
tylerharter, line 164: Do we need to mess with /host/files if we only support mirrors over HTTP?

docs/worker/pip-mirror.md
```diff
@@ -0,0 +1,176 @@
...
+pip's `--no-deps` please ensure that any packages you wish to 
+install have all required dependencies
+within the mirror as well. For more information on how OpenLambda
+installs packages view: [pypi-packages](pypi-packages.md). 
+
+Once you have set up the bandersnatch mirror and are happy with the
+configuration file you can run `bandersnatch mirror` once more to 
+begin installation. This may take several minutes.
+
+### Nginx Support
```
tylerharter, line 52: Are there shorter directions we can give for nginx?  E.g., is there a one-liner we can use to start a container with nginx?  https://hub.docker.com/_/nginx

docs/worker/pip-mirror.md
```diff
@@ -0,0 +1,176 @@
...
+file you can set different settings for the mirror.
+
+Furthermore, you can decide the backend for the bandersnatch mirror.
+This can be a docker container, web server (such as nginx), or a filesystem.
+
+OpenLambda only supports a docker container or web server.
+
+For documentation on how to set up the mirror use: [bandersnatch docs](https://bandersnatch.readthedocs.io/en/latest/mirror_configuration.html)
+
+It is important to note that if you do not use an allowlist or
```
tylerharter, line 33: Should we give an example of the allowlist?

docs/worker/pip-mirror.md
```diff
@@ -0,0 +1,176 @@
+# Installing a Pip Mirror
+
+When running python based lambdas in OpenLambda packages
+are installed with pip. By default, pip installs packages
+through an API to pypi.org, which cannot always be trusted.
+
+Pypi.org cannot guarantee that every indexed package is safe,
+thus a local mirrored index for pip may be beneficial.
```
tylerharter, line 8: Maybe you can describe two advantages: the performance advantage and the security advantage (an allowlist).

docs/worker/pip-mirror.md
```diff
@@ -0,0 +1,176 @@
+# Installing a Pip Mirror
+
+When running python based lambdas in OpenLambda packages
```
tylerharter, line 3: Hyphenate: "Python-based lambdas"

tylerharter (changes requested):
Thanks for the hard work on this!  I really appreciate the effort put into documentation as well.

src/worker/lambda/handlerPuller.go
Siggyv: Only a small change just to copy the requirements.txt into the target dir, because without copying it any installs with requirements.txt would be ignored as it would not be seen by packagepuller.go

src/worker/embedded/packagePullerInstaller.py
Siggyv: You are right there will always be the key provided. If the user doesnt have a mirror listed in the config then the string will be empty, this change will be reflected in the next commit

src/worker/embedded/packagePullerInstaller.py
Siggyv: Totally agree. Let me know if the code in the next commit looks better.

src/worker/embedded/packagePullerInstaller.py
Siggyv: I changed it so that the user enters it in the config.

docs/worker/pip-mirror.md
Siggyv: When I use OL I see it in the console, would a logging system be better? I can create a logging system and have it write to a file? I think its really more so valuable for the user if they are using a mirror since it is not guaranteed the index will have the package they are requesting, while with pypi it is more likely (granted spelling etc is correct). I can add it to the base install_cmds

docs/worker/pip-mirror.md
Siggyv: That part of the readme is in reference to the allocation of the unpacked .whl/ .tar.gz files after installing from the index which are then installed there. So its not restricted to the type of mirror, say a mirror hosted locally would run into the same issues.

docs/worker/pip-mirror.md
Siggyv: Yep!

docs/worker/pip-mirror.md
Siggyv: I think that would be helpful. I have included one in the next commit.

PR #264, opened 2024-12-23
Updated support for Docker sandboxes V2

nboeder7:
This pull request is a rewrite of #262. The primary change is that dockerPool.go now creates containers directly from the original Docker image rather than relying on directory binds on the flattened image. Additionally, this PR adds more changes to testing and CI workflows to better support Docker. The three concerns outlined in PR #262 remain unresolved.

### DIFF
--- a/src/common/config.go
+++ b/src/common/config.go
@@ -64,6 +64,9 @@ type Config struct {
 	// which OCI implementation to use for the docker sandbox (e.g., runc or runsc)
 	Docker_runtime string `json:"docker_runtime"`
 
+	// name of the image used for Docker containers
+	Docker_base_image string `json:"docker_base_image"`
+
 	Limits   LimitsConfig   `json:"limits"`
 	Features FeaturesConfig `json:"features"`
 	Trace    TraceConfig    `json:"trace"`
@@ -131,7 +134,7 @@ type LimitsConfig struct {
 
 // Choose reasonable defaults for a worker deployment (based on memory capacity).
 // olPath need not exist (it is used to determine default paths for registry, etc).
-func LoadDefaults(olPath string) error {
+func LoadDefaults(olPath string, dockerImg string) error {
 	workerDir := filepath.Join(olPath, "worker")
 	registryDir := filepath.Join(olPath, "registry")
 	baseImgDir := filepath.Join(olPath, "lambda")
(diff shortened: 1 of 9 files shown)


src/worker/sandbox/dockerPool.go
```diff
@@ -94,19 +105,33 @@ func (pool *DockerPool) Create(parent Sandbox, isLeaf bool, codeDir, scratchDir
 		pkgDirs = append(pkgDirs, "/packages/"+pkg+"/files")
 	}
 
+	// create the container using the specified configuration
+	procLimit := int64(common.Conf.Limits.Procs)
+	swappiness := int64(common.Conf.Limits.Swappiness)
+	cpuPercent := int64(common.Conf.Limits.CPU_percent)
+	var memoryLimit int64
+	if strings.Contains(scratchDir, "scratch") {
```
tylerharter, line 113: Let's use the meta.MemLimitMB to decide the memory limit.  That should be specified at a higher level, like here: https://github.com/open-lambda/open-lambda/blob/main/src/worker/lambda/packages/packagePuller.go#L180.

src/common/config.go
```diff
@@ -64,6 +64,9 @@ type Config struct {
 	// which OCI implementation to use for the docker sandbox (e.g., runc or runsc)
```
tylerharter, line 64: As Config gets more complicated, some fields are only relevant to certain implementations.  E.g., Docker_runtime and Docker_base_image are only used when we use Docker containers.  It feels like we should isolate implementation-specific fields into sub structs.

src/common/config.go
```diff
@@ -131,7 +134,7 @@ type LimitsConfig struct {
 
 // Choose reasonable defaults for a worker deployment (based on memory capacity).
 // olPath need not exist (it is used to determine default paths for registry, etc).
-func LoadDefaults(olPath string) error {
+func LoadDefaults(olPath string, dockerImg string) error {
```
tylerharter, line 137: It feels a bit off that we tell a function called LoadDefaults the specific value for one of the fields.  I think it is better for LoadDefaults to not take dockerImg and hardcode a default for Docker_base_image.  A caller of LoadDefaults can override then override that field in the returned struct if it wishes.

tylerharter (changes requested):
Great work!  It's nice to also see some resource limits being set and to have the better test filtering.

tylerharter (approved):
Thanks, it's great to support Docker again!

PR #267, opened 2024-12-28
Prevent segfault during `ol worker force-cleanup`

kaimast:
It seems like OL does not load the configuration before using it when calling force-cleanup and segfaults on a nullpointer dereference.

I added a test for Conf being null to checkStatus as well, which is where the segfault happened for me. It is probably not the best idea to have common.Conf as a global variable that can be uninitialized, but changing that is a lot of work.

Finally, I ran `go fmt` which also fixed some other formatting issues.

Maybe @tylerharter or @nannoda can let me know if this looks okay?


tylerharter:
Thanks, good catch!

PR #271, opened 2025-02-17
Bump dependencies in most tests

kaimast:


kaimast:
@tylerharter you should take a look if this breaks anything on your side, before I merge this. The aim of this PR is to get rid of most of the security alerts on the repository right now. 

I had to make some more involved changes, because, except for 1.26, all versions of numpy 1.x already reached end of life. This means I had to move most of the tests to numpy 2.x. There is still the pandas-v1 test which relies on numpy 1.26 now, but we should update that too eventually.

tylerharter:
Cool, thanks for fixing this up!  update_requirements.py was a great idea, hopefully that makes it faster in the future. :)

PR #272, opened 2025-02-25
Prototype of kafka trigger

Poovaya:
Launch the worker with Kafka as the trigger set in the config file, and it will process kafka messages from a broker server and topic in a provided config file


src/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/open-lambda/open-lambda/ol
 
-go 1.18
+go 1.21
```
tylerharter, line 3: If we want to bump the version (good idea, btw!), let's do that in a PR by itself.  That PR should also update the versions used by the tests: https://github.com/open-lambda/open-lambda/tree/main/.github/workflows.

We should check the docs, too (though it looks like that is already on 1.21.5: https://github.com/open-lambda/open-lambda/blob/main/docs/worker/getting-started.md).

src/vendor/modules.txt
```diff
@@ -71,6 +71,10 @@ github.com/Microsoft/go-winio/internal/fs
 github.com/Microsoft/go-winio/internal/socket
 github.com/Microsoft/go-winio/internal/stringbuffer
 github.com/Microsoft/go-winio/pkg/guid
+# github.com/confluentinc/confluent-kafka-go v1.9.2
```
tylerharter, line 74: We've been vendorizing packages (https://github.com/open-lambda/open-lambda/tree/main/src/vendor) so that we have everything necessary to build this project in our own repo.  Copies of everything makes our repo bigger, but less vulnerable to dependencies going away.

src/worker/event/lambdaServer.go
```diff
@@ -16,6 +20,29 @@ type LambdaServer struct {
 	lambdaMgr *lambda.LambdaMgr
 }
 
+// Converts a kafka message to a HTTP request, since the LambdaServer Invocation and everything else relies on
```
tylerharter, line 23: We should do some refactoring of the code in the events directory.  Each event type should be its own file (at this point, one for HTTP and one for Kafka).

src/worker/event/lambdaServer.go
```diff
@@ -67,6 +94,56 @@ func (s *LambdaServer) RunLambda(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (s *LambdaServer) LaunchKafkaConsumer() {
+	// olPath := common.Conf.Worker_dir
+	// kafkaConfigPath := common.Conf.KafkaConfigPath
+	// err := common.LoadKafkaConf(filepath.Join(olPath, kafkaConfigPath))
+	// TODO: Get the path to the ol dir better
+	err := common.LoadKafkaConf("/home/vboxuser/Documents/openl/open-lambda/myworker_k5/kafkaConfig.json")
```
tylerharter, line 102: Please don't hardcode paths that will only work on your machine.

src/common/config.go
```diff
@@ -14,6 +14,17 @@ import (
 
 // Configuration is stored globally here
 var Conf *Config
+var KafkaConf *KafkaConfig
+
+// KafkaConfig represents the configuration for a Kafka listener
+type KafkaConfig struct {
```
tylerharter, line 20: We should add support for some per-lambda configs.  Different functions might subscribe to different clusters/topics.

src/vendor/modules.txt
tylerharter: If vendoring adds a bunch of files (probably will), let's see if we can do that in its own PR.

src/worker/event/lambdaServer.go
```diff
@@ -89,14 +166,18 @@ func NewLambdaServer() (*LambdaServer, error) {
...
 	}
-
-	log.Printf("Setups Handlers")
-	port := fmt.Sprintf(":%s", common.Conf.Worker_port)
-	http.HandleFunc(RUN_PATH, server.RunLambda)
-	http.HandleFunc(DEBUG_PATH, server.Debug)
-
-	log.Printf("Execute handler by POSTing to localhost%s%s%s\n", port, RUN_PATH, "<lambda>")
-	log.Printf("Get status by sending request to localhost%s%s\n", port, STATUS_PATH)
+	switch common.Conf.Trigger {
```
tylerharter, line 169: Let's not be required to do one type or the other at the granularity of a worker.  I'm imagining a function config can indicate it should be triggered by Kafka, HTTP, or both.

tylerharter (changes requested):
Great progress!

PR #274, opened 2025-02-27
added config files

Mandukhai-Alimaa:
Added some example config file (echo.yaml) and default config file in case the user has not uploaded config file for the specific function (default_config.yaml).


test-registry/function_configs_yaml/default_config.yaml
```diff
@@ -0,0 +1,5 @@
+name: default_config
```
tylerharter, line 1: Let's add any default settings to a subfield in Config rather than create an additional worker-wide config file: https://github.com/open-lambda/open-lambda/blob/main/src/common/config.go#L19

test-registry/function_configs_yaml/echo.yaml
```diff
@@ -0,0 +1,16 @@
+name: echo
```
tylerharter, line 1: In one organization, the whole function (with all files related) are in a single .tar.gz: https://github.com/open-lambda/open-lambda/blob/main/src/worker/lambda/handlerPuller.go#L110.  So let's not have a config in a different directory.  It should be in the same directory where the function code is.

Let's use the same config name for every lambda.  Maybe ol.yaml?  This would be similar to how we check to see if there is a requirements.txt file: https://github.com/open-lambda/open-lambda/blob/main/src/worker/lambda/lambdaFunction.go#L77.

tylerharter (changes requested):
Thanks!

tylerharter (changes requested):
Getting closer!

src/common/lambdaConfig.go
```diff
@@ -0,0 +1,92 @@
...
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"gopkg.in/yaml.v3"
+)
+
+var LambdaConf *LambdaConfig
```
tylerharter, line 12: I global makes sense for the worker config (because there is one per process), but we will want different configs for each lambda function.

Currently, each function has sandbox metadata (that gets populated based on requirements.txt) here: https://github.com/open-lambda/open-lambda/blob/main/src/worker/lambda/lambdaFunction.go#L30.  Perhaps we should have a FunctionMeta that includes both a SandboxMeta and a LambdaConfig?  Or something like that?

This function (https://github.com/open-lambda/open-lambda/blob/main/src/worker/lambda/lambdaFunction.go#L71) could possibly load both: your LambdaConfig (from an ol.yaml) and SandboxMeta (from a SandboxMeta).

src/common/lambdaConfig.go
```diff
@@ -0,0 +1,92 @@
...
+
+type HTTPTrigger struct {
+	Method string `yaml:"method"` // HTTP method (e.g., GET, POST)
```
tylerharter, line 15: Can we have a "*" option (for all methods) and default to that?

tylerharter: Once we add this config, we should allow HTTP requests (or not) based on the HTTP trigger config.

The general HTTP handler is RunLambda, and it gets a function from the Lambda Manager here: https://github.com/open-lambda/open-lambda/blob/main/src/worker/event/lambdaServer.go#L62

Before invoking it, we should check the config for that lambda: does it have an HTTP trigger corresponding to the Request Method?  https://pkg.go.dev/net/http#Request.  If not, we should return an error.

src/common/lambdaConfig.go
```diff
@@ -0,0 +1,92 @@
...
+
+type CronTrigger struct {
+	Schedule string `yaml:"schedule"` // Cron schedule (e.g., "*/5 * * * *")
+}
+
+// TODO: add KafkaTrigger struct
+
+// LambdaConfig defines the overall configuration for the lambda function.
+type LambdaConfig struct {
+	HTTPTriggers []HTTPTrigger `yaml:"http,omitempty"` // List of HTTP triggers
```
tylerharter, line 26: We should probably group these into a "triggers" subcategory.  Because we will also have another subcategory related to perf soon (per-lambda memory limit, max runtime, etc).

tylerharter (changes requested):
Thanks!  Suggested some reorg about where to store the config (shouldn't be global because different functions have different configs).

src/worker/event/lambdaServer.go
```diff
@@ -62,12 +62,15 @@ func (s *LambdaServer) RunLambda(w http.ResponseWriter, r *http.Request) {
 			// s.lambdaMgr.Get(img).Invoke(w, r)
 			lambda := s.lambdaMgr.Get(img)
```
Mandukhai-Alimaa, line 63: Tyler, in the Get function, a Task goroutine is started. This Task goroutine calls pullHandlerIfStale, which in turn calls parseMeta to parse the ol.yaml and requirements.txt files. After calling Get and obtaining the LambdaFunc object, I need to check if the HTTP method is a valid trigger. However, I’m running into a race condition: when I get the LambdaFunc back, there’s no guarantee that the Meta field has been fully populated by the Task goroutine. There are a few ways to solve this problem, but I’m not sure which one to use:

1. I can initialize the meta in the get method synchronously with dummy lambda.meta.sandbox value, and load the lambda.meta.Config and let the Task go routine fill in the the lambda.meta.sandbox later
2. I can use a channel to signal completion of the meta field in the lambdaFunc struct then when its ready, I check if httpmethod is the valid trigger then invoke the function
3. I can also use waitgroup to wait for the goroutine to finish, then proceed.

The first way would be simplest way to do it but it not necessarily a clean way to do it. The other ways involve waiting for the meta to be populated or waiting for the goroutines to finish which will slow down the process. Any ideas?

src/worker/lambda/lambdaFunction.go
Poovaya, line 1: I'm not sure if the per function LambdaConfig should be used in this file. The place where the triggers should be handled, i.e., the http server for http triggers is launched, or where he Kafka consumer is created, is lambdaServer.go.

I'm refactoring the code to have each trigger in separate files, but lambdaServer.go will be where the config is checked to decide what triggers the worker should respond to. Once that happens, and one of the triggers is activated (by curl or kafka, or maybe even directly by the boss in the case of cron), this file, lambdaFunction.go, fetches code and dispatches to a lambda Instance. 

The check for whether a HTTP method is allowed should also be in the (to be created) triggers package

Poovaya (commented):
Added some thoughts on the design for LambdaConfig

src/worker/lambda/lambdaFunction.go
```diff
@@ -246,6 +266,14 @@ func (f *LambdaFunc) Task() {
 				continue
 			}
 
+			// Check if the HTTP method is valid
+			if !f.Meta.Config.IsHTTPMethodAllowed(req.r.Method) {
+				req.w.WriteHeader(http.StatusMethodNotAllowed)
+				req.w.Write([]byte("HTTP method not allowed\n"))
```
tylerharter, line 272: Can we say what method was sent and what is allowed?

tylerharter (approved):
Great work!

tylerharter (approved):

PR #275, opened 2025-03-03
add tests for lambda config

Mandukhai-Alimaa:
Added test function in test.py and added a folder with ol.yaml and lambda function.


src/worker/lambda/lambdaFunction.go
```diff
@@ -279,6 +279,8 @@ func (f *LambdaFunc) Task() {
 				continue
 			}
 
+			log.Printf("Method is allowed based on the lambdaConfig")
```
tylerharter, line 282: Let's not log this

src/worker/lambda/handlerPuller.go
```diff
@@ -42,27 +42,26 @@ func Copy(src, dest string) error {
...
 		return filepath.WalkDir(src, func(path string, d fs.DirEntry, err error) error {
-		    if err != nil {
-			return err
-		    }
-
-		    relPath, err := filepath.Rel(src, path)
-		    if err != nil {
-			return err
-		    }
+			if err != nil {
```
tylerharter, line 47: what are these changes about?  are you fixing some sort of bug?  If so, perhaps a separate PR makes sense

src/common/lambdaConfig.go
```diff
@@ -90,6 +90,7 @@ func LoadLambdaConfig(codeDir string) (*LambdaConfig, error) {
 		return nil, fmt.Errorf("failed to parse YAML file: %v", err)
 	}
 
+	fmt.Printf("Parsed config: %+v\n", config)
```
tylerharter, line 93: Let's remove this print

min-image/runtimes/python/server.py
```diff
@@ -28,19 +28,44 @@ def web_server():
...
-        def post(self):
+        # TODO: we should consider how are the different requests used in the context of different applications and functions
+        # and consider what does the validations should look like for example, should we allow POST requests with no payload etc.
+        def handle_request(self):
             try:
                 data = self.request.body
-                try :
-                    event = json.loads(data)
+                try:
+                    event = json.loads(data) if data else None # parse the data if its there, if not, set to None
```
tylerharter, line 37: I think the code is pretty clear, and the comment actually makes it harder to read

tylerharter (changes requested):
Thanks!

tylerharter (approved):
Great work!

PR #276, opened 2025-03-05
Updated config.go to use YAML instead of JSON

anushkaShekar:
I imported the "gopkg.in/yaml.v3" package and used the yaml.Marshal and yaml.Unmarshal functions in place of the json.Marshal and json.Unmarshal functions. Also, I changed the JSON struct tags to YAML struct tags.

### DIFF
--- a/src/common/config.go
+++ b/src/common/config.go
@@ -1,7 +1,7 @@
 package common
 
 import (
-	"encoding/json"
+	//"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 	"syscall"
 
+	"gopkg.in/yaml.v3"
 	"github.com/urfave/cli/v2"
 )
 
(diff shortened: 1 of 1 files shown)


anushkaShekar:
@tylerharter, I've updated the config.go file to use YAML instead of JSON. Please let me know if I should make any additional changes. Thanks!

src/common/config.go
```diff
@@ -259,36 +260,36 @@ func checkConf() error {
 	return nil
 }
 
-// SandboxConfJson marshals the Sandbox_config of the Config into a JSON string.
+// SandboxConfJson marshals the Sandbox_config of the Config into a YAML string.
```
tylerharter, line 263: Please update function name/comment

src/common/config.go
```diff
@@ -1,14 +1,15 @@
 package common
 
 import (
-	"encoding/json"
+	//"encoding/json"
```
tylerharter, line 4: Better to just delete than leave old code.

tylerharter (changes requested):
Good start!  I think you'll have to update some other things as well:

 - tests, like here: https://github.com/open-lambda/open-lambda/blob/main/scripts/helper/__init__.py
 - config.json => config.yaml, places like this: https://github.com/open-lambda/open-lambda/blob/main/src/worker/commands.go

src/common/config.go
anushkaShekar: Fixed this!

src/common/config.go
anushkaShekar: Fixed this too!

tylerharter:
Moved to @297.

PR #277, opened 2025-03-05
added lambda configuration documentation

Mandukhai-Alimaa:
added documentation for the lambda config and linked it in README

### DIFF
--- a/docs/worker/lambda-config.md
+++ b/docs/worker/lambda-config.md
@@ -1,0 +1,49 @@
+# Lambda Configuration Guide
+
+This document provides an overview of per-lambda configuration in Open Lambda, detailing trigger types, configuration options, and usage.
+
+## 1. Overview
+Each lambda function in Open Lambda can be configured using an `ol.yaml` file located in the function's directory. This configuration file defines how the function is triggered and any specific execution parameters.
+
+## 2. Configuration Structure
+A typical `ol.yaml` file follows this structure:
+
+```yaml
+triggers:
+  http:
+    - method: PUT
+    - method: PATCH
+```
+
+## 3. Trigger Types
+Open Lambda only supports HTTP trigger for now but future development plans include supporting other trigger types.
+
+### a. HTTP Triggers
+Defines which HTTP methods can be used to invoke the lambda.
+
+Example:
+```yaml
+triggers:
+  http:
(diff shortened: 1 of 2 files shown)


docs/worker/lambda-config.md
```diff
@@ -0,0 +1,49 @@
+# Lambda Configuration Guide
+
+This document provides an overview of per-lambda configuration in Open Lambda, detailing trigger types, configuration options, and usage.
+
+## 1. Overview
+Each lambda function in Open Lambda can be configured using an `ol.yaml` file located in the function's directory. This configuration file defines how the function is triggered and any specific execution parameters.
```
tylerharter, line 6: No space in OpenLambda

tylerharter: "can [optionally] be configured"

docs/worker/lambda-config.md
```diff
@@ -0,0 +1,49 @@
...
+
+```yaml
+triggers:
+  http:
+    - method: PUT
+    - method: PATCH
+```
+
+## 3. Trigger Types
+Open Lambda only supports HTTP trigger for now but future development plans include supporting other trigger types.
```
tylerharter, line 19: Comma before "but"

docs/worker/lambda-config.md
```diff
@@ -0,0 +1,49 @@
...
+Create an `ol.yaml` file inside the lambda function directory with the desired configuration.
+
+### b. Deploy and Run
+Ensure the Open Lambda framework is set up and execute your lambda as per the defined triggers.
+
+## 5. Validations
+- HTTP triggers must specify valid HTTP methods (GET, POST, PUT, DELETE, etc.).
+- If no triggers are specified or no configuration file exists in the lambda function directory, Open Lambda will apply default behavior allowing all HTTP methods.
+
+## 7. Conclusion
```
tylerharter, line 44: I think the conclusion and everything below can be removed.  I don't think we want to claim to much about ensuring security unless we're more explicit about what that entails.  We also should not refer to the docs when this is the best documentation right here.

tylerharter (changes requested):
Great documentation, thanks!

tylerharter (approved):

PR #281, opened 2025-03-13
Setup dev environment

Mandukhai-Alimaa:
setup local worker as a process with the boss.


src/boss/cloudvm/local_worker.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+func (_ *LocalWorkerPoolPlatform) NewWorker(workerId string) *Worker {
+	return &Worker{
+		workerId: workerId,
+		workerIp: "",
+	}
+}
+
+func (_ *LocalWorkerPoolPlatform) CreateInstance(worker *Worker) {
```
tylerharter, line 32: Maybe we should leave it to a separate PR (and have a TODO for now), but it seems like CreateInstance and DeleteInstance should be returning errors instead of panic'ing or suppressing them.

src/boss/cloudvm/local_worker.go
```diff
@@ -0,0 +1,99 @@
...
+	log.Printf("Creating new local worker: %s\n", worker.workerId)
+
+	// Initialize the worker directory if it doesn't exist
+	initCmd := exec.Command("./ol", "worker", "init", "-p", worker.workerId, "-i", "ol-min")
```
tylerharter, line 36: Leave "TODO" to fix the "ol-min" hardcoding at some point.

src/boss/cloudvm/local_worker.go
```diff
@@ -0,0 +1,99 @@
...
+	initCmd.Stderr = os.Stderr // Capture stderr
+	if err := initCmd.Run(); err != nil {
+		log.Printf("Failed to initialize worker %s: %v\n", worker.workerId, err)
+		panic(err)
+	}
+
+	// Get the executable directory
+	execPath, err := os.Executable()
```
tylerharter, line 44: Left a note below, but don't assume it is relative to the binary.

src/boss/cloudvm/local_worker.go
tylerharter, line 36: The reasons are a little strange, but it will be much faster if we start the worker detached: https://github.com/open-lambda/open-lambda/blob/main/src/worker/commands.go#L93

src/boss/boss.go
```diff
@@ -7,11 +7,12 @@ import (
 	"log"
 	"net/http"
 	"os"
-	"strconv"
```
tylerharter, line 10: Whitespace/formatting improvements are nice, but unless we're doing other changes in the file, let's do these as it's own PR so we don't clutter the logical changes.  You can selectively "git add" files to only include those that matter.

tylerharter (changes requested):
Good work!

tylerharter (changes requested):
Thanks!

src/boss/cloudvm/local_worker.go
```diff
@@ -0,0 +1,89 @@
...
+
+func NewLocalWorkerPool() *WorkerPool {
+	return &WorkerPool{
+		WorkerPoolPlatform: &LocalWorkerPoolPlatform{},
+	}
+}
+
+func (_ *LocalWorkerPoolPlatform) NewWorker(workerId string) *Worker {
+	return &Worker{
+		workerId: workerId,
```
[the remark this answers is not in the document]
Mandukhai-Alimaa: I think it is incrementing the workerId it here and passing it to NewWorker. https://github.com/open-lambda/open-lambda/blob/main/src/boss/cloudvm/worker.go#L109

tylerharter (approved):

PR #287, opened 2025-03-23
platform function returning errors

Mandukhai-Alimaa:
return errors from platform functions

### DIFF
--- a/src/boss/cloudvm/gcp_worker.go
+++ b/src/boss/cloudvm/gcp_worker.go
@@ -87,7 +87,7 @@ func (_ *GcpWorkerPool) NewWorker(workerId string) *Worker {
 	}
 }
 
-func (pool *GcpWorkerPool) CreateInstance(worker *Worker) {
+func (pool *GcpWorkerPool) CreateInstance(worker *Worker) error {
 	client := pool.client
 	fmt.Printf("creating new VM from snapshot\n")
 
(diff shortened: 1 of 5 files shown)


src/boss/cloudvm/local_worker.go
```diff
@@ -28,20 +28,21 @@ func (_ *LocalWorkerPoolPlatform) NewWorker(workerId string) *Worker {
 	}
 }
 
-func (_ *LocalWorkerPoolPlatform) CreateInstance(worker *Worker) {
+func (_ *LocalWorkerPoolPlatform) CreateInstance(worker *Worker) error {
 	log.Printf("Creating new local worker: %s\n", worker.workerId)
 
 	// Initialize the worker directory if it doesn't exist
 	initCmd := exec.Command("./ol", "worker", "init", "-p", worker.workerId, "-i", "ol-min") // TODO fix the "ol-min hardcoding"
 	initCmd.Stderr = os.Stderr
```
tylerharter, line 36: Writes to stdout/std err should usually be lock-protected so we don't get jumbled output (which is possible when we have two different processes writing to it).  Let's not fix it now, let's just leave a TODO.

src/boss/cloudvm/gcp_worker.go
```diff
@@ -97,25 +97,29 @@ func (pool *GcpWorkerPool) CreateInstance(worker *Worker) {
...
 
 	worker.host = lookup[worker.workerId]
 
 	worker.runCmd("./ol worker up -d")
+
+	return nil
 }
 
-func (pool *GcpWorkerPool) DeleteInstance(worker *Worker) {
+func (pool *GcpWorkerPool) DeleteInstance(worker *Worker) error {
```
tylerharter, line 115: It seems this never returns an error.  That's actually pretty worrisome because if we don't delete an instance, but the boss thinks it is deleted, the boss could created too many workers, creating a big bill.  I know it's not a bug in your code, so let's leave a TODO to fix this.

tylerharter (approved):
There are a few issues to resolve in other PRs.  So lets leave some TODOs and get this in.  Great work!

tylerharter (approved):
Great work!

PR #288, opened 2025-03-24
increment port number

Mandukhai-Alimaa:


tylerharter (changes requested):
Won't the worker config always be the template port config + 1?  I think you need to load Worker_Starting_Port somewhere to a new variable that you increment each time before writing a worker config.

src/boss/cloudvm/local_platform_config.go
```diff
@@ -4,13 +4,19 @@ import (
...
 	"path/filepath"
+	"strconv"
 )
 
 type LocalPlatConfig struct {
 	Worker_Starting_Port           string `json:"worker_starting_port"`
 	Path_To_Worker_Config_Template string `json:"path_to_worker_config_template"`
 }
 
+var (
```
tylerharter, line 15: currentWorkerPort (actually "nextWorkerPort" is probably a better name) would make more sense in our struct (https://github.com/open-lambda/open-lambda/blob/main/src/boss/cloudvm/local_worker.go#L14) rather than as a global variable.  You can get rid of portInitialized while you're at it.

src/boss/cloudvm/local_platform_config.go
```diff
@@ -22,3 +28,14 @@ func GetLocalPlatformConfigDefaults() *LocalPlatConfig {
 		Path_To_Worker_Config_Template: filepath.Join(currPath, "template.json"),
 	}
 }
+
+func GetNextWorkerPort() string {
```
tylerharter, line 32: Instead of a function here, it makes more sense as a method in local_worker.go.

src/boss/cloudvm/worker_config_template.go
```diff
@@ -32,6 +32,15 @@ func LoadWorkerConfigTemplate(templatePath string, workerPath string) error {
 		return fmt.Errorf("failed to load template.json: %v", err)
 	}
 
+	// Patch worker-specific paths
```
tylerharter, line 35: Hmmm, this is a bit of a problem.  Because what if they actually want to specify a different zygotes file, or something like that?

I'm not sure of the best way to handle this.  Thoughts?

One idea: if the field is empty in the template, we take that to mean we're supposed to patch it in.  But if it is not empty, we assume the user meant to specify something, and it should be left alone.

If we do override something, we should probably print something so it is less likely to go unnoticed.  I could see this being a point of annoyance for users who are trying to set these fields.

tylerharter (changes requested):

src/boss/cloudvm/local_worker.go
```diff
@@ -6,18 +6,22 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 
 	"github.com/open-lambda/open-lambda/ol/common"
 )
 
 // WORKER IMPLEMENTATION: LocalWorker
 type LocalWorkerPoolPlatform struct {
```
tylerharter, line 15: I think we need to write thread-safe code because we're supposed to support concurrent creation of instances: https://github.com/open-lambda/open-lambda/blob/main/src/boss/cloudvm/worker_pool.go#L122.  So LocalWorkerPoolPlatform should probably have a lock to protect nextWorkerPort and other state that might change.

src/boss/cloudvm/local_worker.go
```diff
@@ -6,18 +6,22 @@ import (
...
-	// no platform specific attributes
+	nextWorkerPort int
 }
 
 func NewLocalWorkerPool() *WorkerPool {
+	startPort, _ := strconv.Atoi(GetLocalPlatformConfigDefaults().Worker_Starting_Port)
```
tylerharter, line 20: Shouldn't we pull from the actual config (https://github.com/open-lambda/open-lambda/blob/main/src/boss/config.go#L14) instead of the default?

src/boss/cloudvm/worker_config_template.go
```diff
@@ -2,13 +2,14 @@ package cloudvm
...
 	"os"
 	"path/filepath"
 
 	"github.com/open-lambda/open-lambda/ol/common"
 )
 
-func LoadWorkerConfigTemplate(templatePath string, workerPath string) error {
+func LoadWorkerConfigTemplate(templatePath string, workerPath string, workerPort string) error {
 	// Create template.json if it doesn't exist
 	if _, err := os.Stat(templatePath); os.IsNotExist(err) {
```
tylerharter, line 14: Instead of checking whether we need to initialize the template each time, why not do this in NewLocalWorkerPool?

Also, common.LoadConf loads the config to a global variable, which we then change, and without lock protection.  That's risky.  Can we make a version of common.LoadConf (w/ a different name) that returns a worker config instead of updating the global variable?   Then we could save the loaded config template in another field in LocalWorkerPoolPlatform, with full control over it and with locking (if needed).

src/boss/cloudvm/worker_config_template.go
```diff
@@ -32,6 +40,30 @@ func LoadWorkerConfigTemplate(templatePath string, workerPath string) error {
 		return fmt.Errorf("failed to load template.json: %v", err)
 	}
 
+	// Patch fields ONLY if they're empty
+	if common.Conf.Worker_dir == "" {
+		common.Conf.Worker_dir = filepath.Join(workerPath, "worker")
+		log.Printf("Patched Worker_dir: %s", common.Conf.Worker_dir) // TODO: protect it with lock
```
tylerharter, line 46: please see email discussion about Printf/locking

src/boss/cloudvm/worker_config_template.go
```diff
@@ -32,6 +40,30 @@ func LoadWorkerConfigTemplate(templatePath string, workerPath string) error {
 		return fmt.Errorf("failed to load template.json: %v", err)
 	}
 
+	// Patch fields ONLY if they're empty
```
tylerharter, line 43: If they're empty, it makes sense to plug in default values.  But can we pull that from the worker defaults instead of hardcoding that here?  E.g., if we want to change the default import cache to be a different zygotes file, I don't want to have to change the code here and in this file: https://github.com/open-lambda/open-lambda/blob/main/src/common/config.go#L143

src/common/config.go
```diff
@@ -212,6 +212,11 @@ func LoadConf(path string) error {
 		return fmt.Errorf("could not parse config (%v): %v", path, err.Error())
 	}
 
+	// Skip checkConf if this is template.json since it will need patching later
```
tylerharter, line 215: Do we need to have a special case?  I worry about basing it on the file name.

tylerharter (changes requested):
Good progress!

src/boss/cloudvm/worker_config_template.go
```diff
@@ -1,42 +1,49 @@
...
 	// Save the template configuration to the worker's config directory
 	configPath := filepath.Join(workerPath, "config.json")
-	if err := common.SaveConf(configPath); err != nil {
-		return fmt.Errorf("failed to save updated configuration to worker config: %v", err)
-	}
+	return common.ExportConfig(cfgCopy, configPath)
+}
 
-	return nil
+func deepCopyConfig(src *common.Config) *common.Config {
```
tylerharter, line 44: If we don't have pointers to sub structs, is there a need for a deep copy?  I guess there are the "any" fields, but we don't change those, do we?

src/boss/config.go
```diff
@@ -12,12 +12,13 @@ import (
...
-	Boss_port  string             `json:"boss_port"`
-	Worker_Cap int                `json:"worker_cap"`
-	Gcp        *cloudvm.GcpConfig `json:"gcp"`
+	Platform   string                   `json:"platform"`
+	Scaling    string                   `json:"scaling"`
+	API_key    string                   `json:"api_key"`
+	Boss_port  string                   `json:"boss_port"`
+	Worker_Cap int                      `json:"worker_cap"`
+	Gcp        *cloudvm.GcpConfig       `json:"gcp"`
+	Local      *cloudvm.LocalPlatConfig `json:"local"`
```
tylerharter, line 21: Is there a motivation for a having a ptr to the sub configs instead of just values?

src/boss/config.go
```diff
@@ -47,6 +49,7 @@ func LoadConf(path string) error {
 	}
 
 	cloudvm.LoadGcpConfig(Conf.Gcp)
```
tylerharter, line 51: I'm not sure I understand why we need this.  When we unmarshall the config, won't we already get the the sub configs from the file?

src/common/config.go
```diff
@@ -288,7 +364,11 @@ func DumpConfStr() string {
 
 // Save writes the Config as an indented JSON to path with 644 mode.
```
tylerharter, line 365: Shouldn't they be named something like SaveGlobalConfig and SaveConfig?

src/common/config.go
```diff
@@ -137,22 +137,32 @@ type LimitsConfig struct {
 // Choose reasonable defaults for a worker deployment (based on memory capacity).
 // olPath need not exist (it is used to determine default paths for registry, etc).
 func LoadDefaults(olPath string) error {
+	cfg, err := GetDefaultWorkerConfig(olPath)
+	if err != nil {
+		return err
+	}
+
+	Conf = cfg
+	return checkConf()
```
tylerharter, line 146: Perhaps if checkConf fails, we shouldn't update the global variable...

src/common/config.go
```diff
@@ -196,23 +206,89 @@ func LoadDefaults(olPath string) error {
...
 		fmt.Printf("Bad config file (%s):\n%s\n", path, string(configRaw))
-		return fmt.Errorf("could not parse config (%v): %v", path, err.Error())
+		return nil, fmt.Errorf("could not parse config (%v): %v", path, err.Error())
 	}
 
-	return checkConf()
+	return &templateConfig, nil
+}
+
+// this function returns the template confif struct with empty worker specific fields
```
tylerharter, line 239: Why do we need this?  Let's not duplicate these defaults in two places (here, and GetDefaultWorkerConfig).

src/boss/cloudvm/worker_config_template.go
```diff
@@ -1,42 +1,49 @@
...
 )
 
-func LoadWorkerConfigTemplate(templatePath string, workerPath string) error {
-	// Create template.json if it doesn't exist
-	if _, err := os.Stat(templatePath); os.IsNotExist(err) {
-		// If template.json does not exist, load defaults
-		if err := common.LoadDefaults(workerPath); err != nil {
-			return fmt.Errorf("failed to load defaults: %v", err)
-		}
+func SaveTemplateConfToWorkerDir(cfg *common.Config, workerPath string, workerPort string) error {
```
tylerharter, line 11: We're constructing a new config based on 3 pieces of info:
1. the template
2. worker default
3. next port number

This is a bit complex so there should be comments explaining what we're doing.

tylerharter (changes requested):
Usually it's good to be consistent with existing code, but I think GCP handles it a bit strangely.  Let's just do this one right and fixup GCP later.

src/boss/cloudvm/worker_config_template.go
```diff
@@ -1,42 +1,46 @@
...
-		return fmt.Errorf("error checking template.json: %v", err)
+// SaveTemplateConfToWorkerDir constructs a worker-specific config using:
+// 1. The global template config (shared defaults)
+// 2. Worker-specific defaults based on its directory path
+// 3. A unique worker port number
+// The final config is then written to <workerPath>/config.json.
+func SaveTemplateConfToWorkerDir(cfg *common.Config, workerPath string, workerPort string) error {
+	// Copy the config so we can safely mutate it
+	cfgCopy := *cfg
+	defaultCfg, _ := common.GetDefaultWorkerConfig(workerPath)
```
tylerharter, line 18: Perhaps we should read this once from the beginning so we don't need to do it each time?

src/boss/cloudvm/worker_pool.go
```diff
@@ -27,7 +27,7 @@ func NewWorkerPool(platform string, worker_cap int) (*WorkerPool, error) {
 	case platform == "gcp":
 		pool = NewGcpWorkerPool()
 	case platform == "local":
-		pool = NewLocalWorkerPool()
+		pool = NewLocalWorkerPool(local_plat_config)
```
tylerharter, line 30: Why do we need to pass it around if we have it in a per-boss global variable?

src/boss/cloudvm/worker_config_template.go
tylerharter, line 18: Why do we ignore errors returned by this function?

tylerharter (approved):
Great work!

PR #289, opened 2025-03-25
add documentation for starting dev environment

Mandukhai-Alimaa:


docs/worker/setup-dev-env.md
```diff
@@ -0,0 +1,67 @@
...
+- Worker config is loaded from **template.json**
+
+### Scale Up to 2 Workers
+```bash
+curl -X POST http://localhost:5000/scaling/worker_count -d "2"
+```
+- A **second worker** starts on port **6001**
+- Each new worker increments the port number
+
+## 4. Customizing Workers
```
tylerharter, line 51: I guess they could change it in the middle, but I don't think it's something we want to encourage.

docs/worker/setup-dev-env.md
```diff
@@ -0,0 +1,67 @@
...
+
+```bash
+vim template.json
+```
+
+### New Workers Use Updated Config
+When scaling up, **new workers** will use the modified `template.json`.
+(Existing workers keep their original config.)
+
+## 5. Shutting Down
```
tylerharter, line 65: They could be running it in the background and do pkill or something like that.  Better to keep it general and say that when the boss receives a kill signal, it will automatically cleanup the workers as well.

tylerharter (changes requested):
This should probably go with boss documentation, not worker documentation: https://github.com/open-lambda/open-lambda/tree/main/docs/boss.

PR #290, opened 2025-03-28
Lambda store

Mandukhai-Alimaa:
building lambda store


src/boss/lambdaStore.go
```diff
@@ -0,0 +1,300 @@
...
+)
+
+type LambdaStore struct {
+	StorePath    string
+	Lambdas      []LambdaEntry
+	HTTPTriggers []HTTPEntry
+	CronTriggers []CronEntry
+}
+
+type LambdaEntry struct {
```
tylerharter, line 25: Do we need all this?  It feels redundant with your code in common: https://github.com/open-lambda/open-lambda/blob/a8ad2825428b10aaed7d8d6472eae00ac91fbc10/src/common/lambdaConfig.go#L13

src/boss/boss.go
```diff
@@ -117,11 +122,22 @@ func BossMain() (err error) {
 		boss.autoScaler.Launch(boss.workerPool)
 	}
 
+	store, err := NewLambdaStore(Conf.Lambda_Store_Path)
```
tylerharter, line 125: The Boss structure created above should be a place to store all the subsystems of the boss:
pool, lambda store, and autoscaler

src/boss/lambdaStore.go
```diff
@@ -0,0 +1,300 @@
...
+
+	s.unregisterTriggers(functionName)
+	s.removeFromRegistry(functionName)
+	s.addToRegistry(functionName, cfg)
+	s.registerTriggers(functionName, cfg)
+
+	return nil
+}
+
+func (s *LambdaStore) registerTriggers(functionName string, cfg *common.LambdaConfig) {
```
tylerharter, line 197: Let's leave the bodies of registerTriggers and unregisterTriggers as TODOs for this PR.  I think events should be a separate subsystem that the registry interacts with instead of having that logic here.  This can eventually end up in boss/event, mirroring worker/event

src/boss/lambdaStore.go
```diff
@@ -0,0 +1,300 @@
...
+	s.removeFromRegistry(functionName)
+	s.addToRegistry(functionName, cfg)
+	s.registerTriggers(functionName, cfg)
+
+	w.WriteHeader(http.StatusCreated)
+	fmt.Fprintf(w, "Lambda %s uploaded successfully", functionName)
+}
+
+func (s *LambdaStore) DeleteLambda(w http.ResponseWriter, r *http.Request) {
+	raw := strings.TrimPrefix(r.URL.Path, "/lambda/")
```
tylerharter, line 108: It seems that creating/deleting files for functions should be done in addToRegistry and removeFromRegistry so that it is easy to keep on-disk and in-memory state in sync.  Then somebody writing new code won't have to worry about taking multiple steps to update both.

src/boss/lambdastore/store.go
```diff
@@ -0,0 +1,230 @@
...
+	store := &LambdaStore{
+		StorePath: storePath,
+		Lambdas:   make(map[string]*common.LambdaConfig),
+	}
+
+	if err := os.MkdirAll(store.StorePath, 0755); err != nil {
+		return nil, fmt.Errorf("failed to create lambda store directory: %w", err)
+	}
+
+	files, _ := os.ReadDir(store.StorePath)
```
tylerharter, line 37: Let's not ignore errors.

src/boss/lambdastore/store.go
```diff
@@ -0,0 +1,230 @@
...
+func (s *LambdaStore) addToRegistry(name string, body io.Reader) error {
+	tarPath := filepath.Join(s.StorePath, name+".tar.gz")
+
+	tarFile, err := os.Create(tarPath)
+	if err != nil {
+		return fmt.Errorf("failed to create lambda tarball: %w", err)
+	}
+	defer tarFile.Close()
+
+	if _, err := io.Copy(tarFile, body); err != nil {
```
tylerharter, line 186: Let's have a TODO to atomically write this with the help of rename

src/boss/lambdastore/store.go
```diff
@@ -0,0 +1,230 @@
...
+	"strings"
+	"sync"
+
+	"github.com/open-lambda/open-lambda/ol/common"
+)
+
+type LambdaStore struct {
+	StorePath string
+	// mapLock protects concurrent access to the Lambdas map
+	mapLock sync.RWMutex
```
tylerharter, line 21: Let's have a TODO to simplify the locking structure as discussed in person.

tylerharter (approved):
Great work!

PR #296, opened 2025-04-17
Added kafka consumer logic

Poovaya:


src/worker/event/kafkaServer.go
```diff
@@ -0,0 +1,119 @@
...
+			iter := fetches.RecordIter()
+			for !iter.Done() {
+				record := iter.Next()
+				r, err := ConvertKafkaMessageToHTTPRequest(record.Value)
+				if err != nil {
+					log.Printf("%f", err.Error())
+					panic(err)
+				}
+				w := httptest.NewRecorder()
+				kafkaServer.lambdaMgr.Get(functionName).Invoke(w, r)
```
tylerharter, line 115: To minimize overhead, can we send the lambda function a whole batch at once?

src/worker/event/kafkaServer.go
```diff
@@ -0,0 +1,119 @@
...
+	// Set headers
+	req.Header.Set("User-Agent", "Kafka-Consumer-Client")
+	req.Header.Set("Accept", "*/*")
+	req.Header.Set("Content-Length", fmt.Sprintf("%d", len(message)))
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded") // Assuming Kafka messages are JSON
+
+	return req, nil
+}
+
+func KafkaInit(w http.ResponseWriter, r *http.Request) {
```
tylerharter, line 47: Can we refactor this function into two?  One for the HTTP side of things, and one for actually starting the goroutine and consuming messages.

Also, it probably makes sense as methods on KafkaServer, no functions that use a global.

tylerharter (changes requested):
Good first draft!

src/worker/event/server.go
```diff
@@ -16,8 +16,22 @@ import (
 	"syscall"
 
 	"github.com/open-lambda/open-lambda/ol/common"
+	"github.com/open-lambda/open-lambda/ol/worker/lambda"
 )
 
+var lambdaMgr *lambda.LambdaMgr
+var once sync.Once
+
+// LambdaManager is now a singleton, one per worker. This is because lambda manager
```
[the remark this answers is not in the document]
Poovaya: For this PR, what would you recommend? If the worker is supposed to support both Kafka and HTTP/Cron triggers at the same time, this change or something similar would be needed for even bare minimum functionality

PR #297, opened 2025-04-21
Updated these files to use YAML instead of JSON

anushkaShekar:
Files Updated: 
1. config.go
2. commands.go
3. __init__.py

### DIFF
--- a/src/boss/cloudvm/gcp_worker.go
+++ b/src/boss/cloudvm/gcp_worker.go
@@ -107,17 +107,17 @@ func (pool *GcpWorkerPool) CreateInstance(worker *Worker) error {
 
 	worker.host = lookup[worker.workerId]
 
-	worker.runCmd("./ol worker up -d")
+	worker.runCmd("./ol worker up -d") // TODO: check if runCmd fails.
 
 	return nil
 }
 
 func (pool *GcpWorkerPool) DeleteInstance(worker *Worker) error {
 	log.Printf("deleting gcp worker: %s\n", worker.workerId)
-	worker.runCmd("./ol worker down")
+	worker.runCmd("./ol worker down")                                // TODO: check if runCmd fails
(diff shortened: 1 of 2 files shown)


anushkaShekar:
@tylerharter @Mandukhai-Alimaa Updated these files to use YAML instead of JSON

scripts/helper/__init__.py
```diff
@@ -53,27 +53,27 @@ def get_current_config():
 def post(path, data=None):
     ''' Issues a post request to the OL worker '''
     return requests.post(f'http://localhost:5000/{path}',
-                         json.dumps(data), timeout=1.0)
+                         yaml.dumps(data), timeout=1.0)
```
tylerharter, line 56: Is data config, or something else?  Not sure yaml is right here...

src/common/config.go
```diff
@@ -259,36 +259,36 @@ func checkConf() error {
 	return nil
 }
 
-// SandboxConfJson marshals the Sandbox_config of the Config into a JSON string.
-func SandboxConfJson() string {
-	s, err := json.Marshal(Conf.Sandbox_config)
+// SandboxConfYaml marshals the Sandbox_config of the Config into a YAML string.
```
tylerharter, line 262: Is this function used anywhere?  Perhaps we should just delete it.

tylerharter (changes requested):
Good progress!  Please make sure you're only changing the config with JSON => YAML.  We're not trying to replace JSON everywhere, and it seems some code got swept up in that.

It looks like there are some merge changes to fix up.

src/common/config.go
anushkaShekar: No it's not used anywhere else; I will delete it when I push again.

scripts/helper/__init__.py
anushkaShekar: Fixed this! Thank you

PR #299, opened 2025-05-07
modified config.go, commands.go, __init__.py

anushkaShekar:


anushkaShekar:
@tylerharter Modified the config.go, commands.go, and __init__.py files to use yaml instead of json. Please let me know if I should make any additional changes. Thanks.

anushkaShekar:
@tylerharter Installed yaml for the CI tests

Mandukhai-Alimaa:
Hi Anushka, It seems that the the CI test is failing because the config.yaml is not created and populated correctly. I believe there is one more change need to be done in https://github.com/open-lambda/open-lambda/blob/main/src/worker/helpers.go#L147.  It is still initializing config.json but trying to read config.yaml.

src/common/config.go
```diff
@@ -292,40 +292,31 @@ func checkConf(cfg *Config) error {
 	return nil
 }
 
-// SandboxConfJson marshals the Sandbox_config of the Config into a JSON string.
-func SandboxConfJson() string {
```
Mandukhai-Alimaa, line 296: Just a quick check — are we confident that this function isn’t currently used or won’t be needed in the near future? Want to make sure we’re not removing something that might be useful for debugging or config serialization down the line.

Mandukhai-Alimaa (changes requested):
Hi Anushka, It seems that the the CI test is failing because the config.yaml is not created and populated correctly. I believe there is one more change need to be done in https://github.com/open-lambda/open-lambda/blob/main/src/worker/helpers.go#L147. It is still initializing config.json but trying to read config.yaml.

src/common/config.go
anushkaShekar: I will check this

anushkaShekar:
@Mandukhai-Alimaa Hi Mandy. I fixed the line in helpers.go to initialize config.yaml instead of config.json, but the CI tests are still failing (i'm getting this same error : FileNotFoundError: [Errno 2] No such file or directory: 'test-dir/config.yaml'). I tried searching for all instances of config.json and replacing it with config.yaml, but that still didn't work. not sure what i'm missing, so any help would be appreciated!

src/common/config.go
anushkaShekar: Hi, @Mandukhai-Alimaa. I did a grep and it appears that this function is not being used anywhere else in the codebase. Should I delete it ?

anushkaShekar:
Hi, @Mandukhai-Alimaa! I was able to get all the tests to pass, and just committed my changes. Please lmk if I need to make any additional changes at your convenience.

Mandukhai-Alimaa (changes requested):
Nice work on this! Just a small thing — it looks like the default-ol and min-image directories were committed by mistake. These shouldn’t be part of the PR, so please remove them. Thanks!

Also, just double-checking — was the intent to update all three configs (worker, boss, and zygote tree) to YAML? I’m asking since Tyler mentioned:

"Good progress! Please make sure you're only changing the config with JSON => YAML. We're not trying to replace JSON everywhere, and it seems some code got swept up in that."

If the change to all three configs is indeed intended, it looks like the updated CI test still expects boss.json. You can go ahead and update the test to look for boss.yaml instead.

PR #300, opened 2025-05-09
Lambda store v2

Mandukhai-Alimaa:


src/boss/lambdastore/store.go
```diff
@@ -104,14 +99,13 @@ func (s *LambdaStore) DeleteLambda(w http.ResponseWriter, r *http.Request) {
 
 func (s *LambdaStore) ListLambda(w http.ResponseWriter, r *http.Request) {
 	s.mapLock.Lock()
+	defer s.mapLock.Unlock()
```
tylerharter, line 102: Holding a lock for the whole execution is usually a good idea, but I do try to avoid it when doing I/O, so perhaps the previous version was better.

src/boss/lambdastore/store.go
```diff
@@ -188,34 +182,42 @@ func (s *LambdaStore) unregisterTriggers(functionName string) {
...
+	if err := os.Rename(tmpPath, tarPath); err != nil {
+		_ = os.Remove(tmpPath)
+		return fmt.Errorf("failed to rename temp tarball: %w", err)
 	}
 
 	s.mapLock.Lock()
 	defer s.mapLock.Unlock()
 
 	entry, ok := s.Lambdas[name]
 	if !ok {
```
tylerharter, line 219: Errors are good for regular stuff that can happen.  If something should never happen (like, we have a bug that breaks assumptions), panic is appropriate.  However, is there a way to avoid this check?  For example, this could be a method on LambdaEntry instead of LambdaStore, so that there is no way to call it with a missing entry.  You could also get rid of getFuncLock -- better to have a function that gets an entry, then use the lock within that entry.

src/boss/lambdastore/store.go
```diff
@@ -188,34 +182,42 @@ func (s *LambdaStore) unregisterTriggers(functionName string) {
...
-		return fmt.Errorf("failed to create lambda tarball: %w", err)
+		return fmt.Errorf("failed to create temp tarball: %w", err)
 	}
-	defer tarFile.Close()
 
-	// TODO: atomically write tarball by writing to a temp file and renaming
 	if _, err := io.Copy(tarFile, body); err != nil {
-		return fmt.Errorf("failed to write tarball: %w", err)
+		tarFile.Close()
+		_ = os.Remove(tmpPath)
```
tylerharter, line 194: Rather than have os.Remove each time, we can have a defer with an anonymous function that always tries to do Remove when addToRegistry returns.

src/boss/lambdastore/store.go
```diff
@@ -188,34 +182,42 @@ func (s *LambdaStore) unregisterTriggers(functionName string) {
...
+		return fmt.Errorf("failed to write to temp tarball: %w", err)
 	}
 
-	cfg, err := common.ExtractConfigFromTarGz(tarPath)
+	if err := tarFile.Close(); err != nil {
+		_ = os.Remove(tmpPath)
+		return fmt.Errorf("failed to close temp tarball: %w", err)
+	}
+
+	cfg, err := common.ExtractConfigFromTarGz(tmpPath)
```
tylerharter, line 203: I can't remember, but are we always supposed to have a config?  Or is it supposed to behave in some reasonable default way even without?

tylerharter (changes requested):
Good work!

src/boss/lambdastore/store.go
Mandukhai-Alimaa: if there is a config, we are parsing it, if not, we are falling back to default config

src/boss/lambdastore/store.go
```diff
@@ -175,49 +170,52 @@ func (s *LambdaStore) loadConfigAndRegister(functionName string) error {
...
-		}
-		s.Lambdas[name] = entry
+	// Atomically replace the old file
+	if err := os.Rename(tmpPath, tarPath); err != nil {
+		return fmt.Errorf("failed to rename temp tarball: %w", err)
 	}
-	entry.Config = cfg
+
+	lambdaEntry := s.getEntry(name)
+	lambdaEntry.Config = cfg
```
tylerharter, line 218: Do we need to use the entry's lock?

tylerharter (approved):
Feel free to merge after addressing the comment (if it makes sense)

src/boss/lambdastore/store.go
Mandukhai-Alimaa: Tyler, this function assumes the caller already holds the entry lock, so accessing lambdaEntry.Config = cfg should be safe here

PR #301, opened 2025-05-15
Integrate Capabilities in Open-Lambda

m0mosenpai:


min-image/runtimes/python/ol.c
```diff
@@ -388,6 +454,8 @@ static PyMethodDef OlMethods[] = {
                                   {"unshare", (PyCFunction)ol_unshare, METH_NOARGS, "unshare"},
                                   {"fork", (PyCFunction)ol_fork, METH_NOARGS, "fork"},
                                   {"enable_seccomp", (PyCFunction)ol_enable_seccomp, METH_NOARGS, "enable_seccomp"},
+                                  {"modify_cap", (PyCFunction)ol_modify_cap, METH_VARARGS, "modify_cap"},
```
m0mosenpai, line 457: @tylerharter This is still a work in progress. But the basic idea (based on my understanding) was that maybe we can expose a function here to modify/ drop capabilities. I'm still not sure which capabilities should be dropped and which ones should be kept?

min-image/runtimes/python/server.py
```diff
@@ -208,6 +208,7 @@ def main():
             file.write(pid)
             print(f'server.py: joined cgroup, close FD {fd_id}')
 
+    # drop privileges here
```
m0mosenpai, line 211: Following the C code, I was thinking this would be one of the places where we would drop the privileges (either here or in `start_container()` just before forking (?) by invoking `ol.modify_caps(args)` (where args would have a list of capabilities we want to keep/ drop)

min-image/runtimes/python/ol.c
```diff
@@ -8,6 +9,71 @@
...
+
+  caps = cap_get_proc();
+  if (caps == NULL) {
+    rc = -1;
+    goto out;
+  }
+
+  capList[0] = cap;
+  if (cap_set_flag(caps, CAP_EFFECTIVE, 1, capList, setting) == -1) {
+    cap_free(caps);
```
tylerharter, line 27: One of the nice things about a goto is it gives you a single place to clean things up, like this:

if (caps != NULL)
    cap_free(caps);

min-image/runtimes/python/ol.c
```diff
@@ -8,6 +9,71 @@
 #include <sys/types.h>
 #include <sys/wait.h>
 #include <seccomp.h>
+#include <sys/capability.h>
+
+static int modify_cap_impl(int cap, int setting) {
```
tylerharter, line 14: To what extent can we just create Python wrappers around the cap API (can_get_proc, cap_set_flag, etc), and then have the logic using it in Python code only?

min-image/runtimes/python/server.py
tylerharter, line 211: Can we make capability dropping optional, via config, like it is for seccomp?  https://github.com/open-lambda/open-lambda/blob/main/min-image/runtimes/python/server.py#L189

tylerharter (changes requested):
Great start!

min-image/runtimes/python/ol.c
m0mosenpai: @tylerharter I actually didn't what you mean here. I tried to use the goto statement already. Is there an even better way to do it?

min-image/runtimes/python/server.py
m0mosenpai: @tylerharter I added the command line argument for enabling capabilities as well. However, we'll need to pass the flag on to everywhere we drop capabilities - to check it and drop only if it's true. Wondering what would be the best way to do that.

m0mosenpai:
@tylerharter "How about this first? "right before forking a Leaf to run the user code". But is it before, or after forking?"

Taking this scenario first, I'm not sure exactly which one would be better or what the difference in the impact would be - dropping capabilities before vs after?

min-image/runtimes/python/ol.c
tylerharter: You have cap_free in multiples places.  What I mean is that cap_free can be after "out:".  You'll need to check whether it is NULL before freeing there, of course.

min-image/runtimes/python/server.py
```diff
@@ -177,15 +185,16 @@ def main():
 
     global bootstrap_path
 
-    if len(sys.argv) < 2:
-        print("Expected execution: chroot <path_to_root_fs> python3 server.py <path_to_bootstrap.py> [cgroup-count] [enable-seccomp]")
+    if len(sys.argv) < 3:
```
tylerharter, line 188: Let's add a TODO to move to named args (in a future PR) now that we're getting more than a couple.

min-image/runtimes/python/server.py
```diff
@@ -158,6 +158,14 @@ def start_container():
         # works for the process that calls it.
         os._exit(0)
 
+    # drop all capabilities
+    return_val = ol.drop_all_caps();
+    assert return_val == 0
+
+    # set CAP_SYS_ADMIN
+    return_val = ol.modify_caps(21, 1);
```
tylerharter, line 166: Can we avoid hardcoding this?  Anyway we can expose some consts (like CAP_SYS_ADMIN) from our ol module to Python?

tylerharter (changes requested):
Looking good!  Looking forward to seeing the next iteration.

min-image/runtimes/python/server.py
tylerharter: Wait, aren't we always doing this, regardless of cmd line args?

min-image/runtimes/python/server.py
```diff
@@ -177,15 +185,16 @@ def main():
...
     global bootstrap_path
 
-    if len(sys.argv) < 2:
-        print("Expected execution: chroot <path_to_root_fs> python3 server.py <path_to_bootstrap.py> [cgroup-count] [enable-seccomp]")
+    if len(sys.argv) < 3:
+        print("Expected execution: chroot <path_to_root_fs> python3 server.py <path_to_bootstrap.py> [cgroup-count] [enable-seccomp] [enable-capabilities]")
         print("    cgroup-count: number of FDs (starting at 3) that refer to /sys/fs/cgroup/..../cgroup.procs files")
-        print("    enable-seccomp: true/false to enable or disables seccomp filtering")
+        print("    enable-seccomp: true/false to enable or disable seccomp filtering")
+        print("    enable-capabilities: true/false to enable or disable capability dropping")
```
tylerharter, line 192: I wonder if we should call it drop-capabilities or something like that?  I think it is intuitive that "enable-seccomp" is going to restrict you.  But it's less obvious whether "enable-capabilities" is going to give you more capabilities, or turn on logic that limits your capabilities.

min-image/Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM ubuntu:22.04
 
+RUN apt-get autoremove
```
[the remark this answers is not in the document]
m0mosenpai: @tylerharter upon setting up OpenLambda for the first time, the DockerFile errored out due to some dependency or cleanup issue and suggested to run this to remove any conflicting stale package in the instance. I don't remember the exact error now but I adding this should make sure none of that happens

min-image/runtimes/python/server.py
m0mosenpai: @tylerharter Right, for that, I was concerned that I would have to pass that boolean flag down the function calls. `enable-seccomp` bool is only needed in the beginning where it is "enabled" and then that's it. But since we would have to drop privileges at certain points in the code, we would need the `drop-capabilities` bool value to propagate down.

I left it as it is for now for testing while I thought of a good solution for it or if you have any suggestions.

min-image/runtimes/python/server.py
m0mosenpai: That makes sense, I'll change the name

min-image/runtimes/python/ol.c
m0mosenpai: @tylerharter I see. Based on my research, I couldn't find any official Python libraries that would give us control over capabilities. There are some community wrappers available but I'm assuming they use similar C code underneath. I'm also concerned that if we rely on them, we could open a channel for untrustworthy libraries/ packages that aren't well maintained or supported. What are your thoughts on that?

PR #302, opened 2025-05-20
Kafka trigger

Mandukhai-Alimaa:


src/boss/event/kafka_manager.go
```diff
@@ -0,0 +1,140 @@
...
+			continue
+		}
+
+		data, err := json.Marshal(trigger)
+		if err != nil {
+			log.Printf("[KafkaManager] Failed to marshal Kafka trigger for %s: %v", functionName, err)
+			continue
+		}
+
+		req := httptest.NewRequest(http.MethodPost, "/kafka-init/"+functionName, bytes.NewBuffer(data))
```
tylerharter, line 70: I don't know what httptest does, but is what we're doing related to testing?  Why not a regular HTTP request, like here: https://github.com/open-lambda/open-lambda/blob/main/src/worker/lambda/lambdaInstance.go#L145?

src/boss/event/kafka_manager.go
```diff
@@ -0,0 +1,140 @@
...
+	for _, trigger := range triggers {
+		// validate the required fields, make sure it is not empty
+		if len(trigger.Topics) == 0 || len(trigger.Bootstrap_servers) == 0 {
+			log.Printf("[KafkaManager] Skipping empty Kafka trigger for %s", functionName)
+			continue
+		}
+
+		data, err := json.Marshal(trigger)
+		if err != nil {
+			log.Printf("[KafkaManager] Failed to marshal Kafka trigger for %s: %v", functionName, err)
```
tylerharter, line 66: Is it right that we have errors here but the function returns no error?

tylerharter (changes requested):
Great work!

PR #306, opened 2025-06-13
adding per-lambda memory limit files

guptadevagya:
I've tried implementing per-lambda memory limits. This would allow a function to define its own memory limit via a memory_mb key in its ol.yaml file.

My changes are:
- I added a MemoryMaxMB field to the LambdaConfig struct in src/common/lambdaConfig.go.
- I updated the sandbox creation logic in src/worker/sandbox/sockPool.go to use this new value. If a per-lambda limit isn't set, it falls back to the global default as expected.

### DIFF
--- a/src/common/lambdaConfig.go
+++ b/src/common/lambdaConfig.go
@@ -37,7 +37,8 @@ type CronTrigger struct {
 
 // LambdaConfig defines the overall configuration for the lambda function.
 type LambdaConfig struct {
-	Triggers Triggers `yaml:"triggers"` // List of HTTP triggers
+	Triggers    Triggers `yaml:"triggers"`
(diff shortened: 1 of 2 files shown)


src/common/lambdaConfig.go
```diff
@@ -37,7 +37,8 @@ type CronTrigger struct {
 
 // LambdaConfig defines the overall configuration for the lambda function.
 type LambdaConfig struct {
-	Triggers Triggers `yaml:"triggers"` // List of HTTP triggers
+	Triggers    Triggers `yaml:"triggers"`
+	MemoryMaxMB int      `yaml:"memory_mb,omitempty"` // List of HTTP triggers
```
Mandukhai-Alimaa, line 41: the comment here seems to be misplaced

PR #314, opened 2025-06-25
add tar support for lambdas

Mandukhai-Alimaa:


src/main.go
```diff
@@ -323,6 +540,26 @@ OPTIONS:
 	app.EnableBashCompletion = true
 	app.HideVersion = true
 	app.Commands = []*cli.Command{
+		&cli.Command{
+			Name:      "admin",
+			Usage:     "Admin commands for managing lambdas",
+			UsageText: "ol admin <cmd>",
+			Subcommands: []*cli.Command{
```
tylerharter, line 547: let's put admin commands to a new file (package?), like we do for with BenchCommands and WorkerCommands

src/worker/event/server.go
```diff
@@ -200,6 +209,40 @@ func shutdown(pidPath string, server cleanable) {
 	os.Exit(rc)
 }
 
+// RegistryHandler handles registry requests similar to boss, but uploads to common.Conf.Registry
```
tylerharter, line 212: Do we need a separate handler, or the the boss and worker use the same code?

src/worker/lambda/handlerPuller.go
```diff
@@ -113,11 +113,9 @@ func (cp *HandlerPuller) Pull(name string) (rt_type common.RuntimeType, targetDi
 	}
 
 	if cp.isRemote() {
-		// registry type = web
+		// registry type = web - only support tar.gz files
```
tylerharter, line 116: isRemote will go away, right?  Seems unnecessary if the way we handle a distributed deployment will be a shared file system across workers/boss

tylerharter: Just thinking out loud (this could be in the next PR).

src/worker/event/registry.go
```diff
@@ -0,0 +1,157 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/open-lambda/open-lambda/ol/common"
+)
+
+// UploadLambda handles POST /registry/{name} - saves to common.Conf.Registry
+func UploadLambda(w http.ResponseWriter, r *http.Request) {
+	funcName := strings.TrimPrefix(r.URL.Path, REGISTRY_BASE_PATH)
```
tylerharter, line 17: should there be any locking at this level, to prevent concurrent upload of the same file?

docs/worker/getting-started.md
```diff
@@ -133,15 +133,37 @@ start and stop a worker many times without reinitializing.  You can
...
+This command will package the `echo` directory into a `.tar.gz` file and upload it to the worker's registry.
+
+If you initialized a worker with a specific path (e.g., `./ol worker init -p myworker`), you must specify the same path when installing a lambda.
+
+```bash
+./ol admin install -p myworker echo/
+```
+
+If no `-p` flag is specified, the command will default to the worker running on port 5000 using the default config.
+
```
tylerharter, line 166: Should we say something about how they can use the examples we provide?

tylerharter (approved):
Looking good!

src/worker/lambda/handlerPuller.go
Mandukhai-Alimaa: I will do this in the next PR.

PR #322, opened 2025-07-15
use gocloud to support distributed lambda_store

Mandukhai-Alimaa:


src/boss/lambdastore/store.go
```diff
@@ -264,8 +296,8 @@ func (s *LambdaStore) removeFromRegistry(funcName string) error {
 
 	// Background deletion
 	go func() {
-		if err := os.Remove(trashPath); err != nil {
-			log.Printf("warning: failed to remove %s from trash: %v", trashPath, err)
+		if err := s.bucket.Delete(context.Background(), funcName+".tar.gz"); err != nil {
```
tylerharter, line 299: It seems like we hardcode .tar.gz in a lot of places.  Might be a pain later if we use a different format.  At the least, how about a const for now?

src/boss/lambdastore/store.go
```diff
@@ -277,16 +309,17 @@ func (s *LambdaStore) getConfig(funcName string) (*common.LambdaConfig, error) {
...
-		}
-		return nil, fmt.Errorf("failed to extract config: %w", err)
+	// Return cached config if available
+	if lambdaEntry.Config != nil {
+		return lambdaEntry.Config, nil
+	}
+
+	// If not cached, try to load from blob storage
+	if err := s.loadConfigAndRegister(funcName); err != nil {
+		return nil, fmt.Errorf("lambda %q not found", funcName)
```
tylerharter, line 319: How do we know the err is not found, and not something else?  Probably better to propagate with %w.

src/common/config.go
```diff
@@ -175,7 +175,7 @@ func GetDefaultWorkerConfig(olPath string) (*Config, error) {
 		Server_mode:       "lambda",
 		Worker_url:        "localhost",
 		Worker_port:       "5000",
-		Registry:          registryDir,
+		Registry:          "file://" + registryDir,
```
tylerharter, line 178: Should we have a comment about why the protocol prefix is necessary?

src/worker/event/server.go
```diff
@@ -278,10 +278,11 @@ func Main() (err error) {
 	http.HandleFunc(PPROF_CPU_STOP_PATH, PprofCpuStop)
 
 	// Initialize LambdaStore for registry
+	log.Printf("Worker: Initializing LambdaStore with Registry = %s", common.Conf.Registry)
```
tylerharter, line 281: Let's have quotes around the %s, so it is obvious the problem if empty.

src/worker/event/server.go
```diff
@@ -278,10 +278,11 @@ func Main() (err error) {
...
 	lambdaStore, err = lambdastore.NewLambdaStore(common.Conf.Registry, nil)
 	if err != nil {
 		os.Remove(pidPath)
-		return fmt.Errorf("failed to initialize lambda store: %v", err)
+		return fmt.Errorf("failed to initialize lambda store at %s: %v", common.Conf.Registry, err)
```
tylerharter, line 285: %v => %w

src/worker/lambda/handlerPuller.go
```diff
@@ -1,294 +1,154 @@
...
-			rt_type, targetDir, err = cp.pullRemoteFile(urls[i], name)
-			if err == nil {
-				return rt_type, targetDir, nil
-			} else if err != errNotFound404 {
-				// 404 is OK, because we just go on to check the next URLs
-				return rt_type, "", err
-			}
+	attrs, err := cp.bucket.Attributes(context.Background(), key)
+	if err == nil {
+		version := attrs.ModTime.String()
```
tylerharter, line 81: Why convert the ModTime to a string, instead of just storing the version as a timestamp?  Do you envision it being represented differently at some point?

PR #325, opened 2025-07-30
change snapshot to include worker template and set the registry path based on the template 

Mandukhai-Alimaa:


src/boss/cloudvm/gcp_worker.go
```diff
@@ -64,6 +68,15 @@ func NewGcpWorkerPool() *WorkerPool {
 	}
 	fmt.Printf("Instance: %s\n", instance)
 
+	fmt.Printf("STEP 2a: prepare snapshot with GCS lambda store config\n")
+	if err := createGcsTemplate(); err != nil {
+		panic(fmt.Errorf("failed to create GCS template.json: %v", err))
```
tylerharter, line 73: %w

src/boss/cloudvm/gcp_worker.go
```diff
@@ -64,6 +68,15 @@ func NewGcpWorkerPool() *WorkerPool {
...
+	}
+
+	// Brief pause to ensure filesystem operations are fully committed before snapshot
```
tylerharter, line 76: What if the change is buffered for longer that 1s?  Can we force it to disk with a sync?  https://pkg.go.dev/os#File.Sync

src/common/config.go
```diff
@@ -305,6 +355,64 @@ func SandboxConfJson() string {
...
+	tempFile = nil // Mark as closed to avoid double-close in defer
+	
+	// Atomic rename - this is the key operation that prevents corruption
+	if err := os.Rename(tempPath, filePath); err != nil {
+		return fmt.Errorf("failed to rename temp file: %v", err)
+	}
+	
+	// Sync the directory to ensure the rename is persisted before snapshot
+	dirFile, err := os.Open(filepath.Dir(filePath))
+	if err == nil {
```
tylerharter, line 407: Shouldn't we check if there is an error?

src/boss/cloudvm/gcp_worker.go
tylerharter: Ah, I see you do sync below.  Was it not working already, without the sleep?  We need to understand this a bit better (I don't trust sleep to robustly solve any problem remaining).

src/common/config.go
```diff
@@ -305,6 +355,64 @@ func SandboxConfJson() string {
 	return string(s)
 }
 
+// SaveConfigAtomic saves a config to a file atomically to prevent race conditions
+// This is used for template.json creation to avoid corruption during concurrent access
```
tylerharter, line 359: Do we actually have concurrent access, or is the real concern that the data is not flushed to disk before the disk snapshot begins?

Is there any downside to just having a regular SaveConfig that always takes special care with sync/rename?

src/common/config.go
```diff
@@ -305,6 +355,64 @@ func SandboxConfJson() string {
...
+func SaveConfigAtomic(cfg *Config, filePath string) error {
+	// Create temp file in same directory as target file
+	dir := filepath.Dir(filePath)
+	tempFile, err := os.CreateTemp(dir, filepath.Base(filePath)+".tmp.*")
+	if err != nil {
+		return fmt.Errorf("failed to create temp file: %v", err)
```
tylerharter, line 365: %w

src/common/config.go
```diff
@@ -305,6 +355,64 @@ func SandboxConfJson() string {
 	return string(s)
 }
 
+// SaveConfigAtomic saves a config to a file atomically to prevent race conditions
+// This is used for template.json creation to avoid corruption during concurrent access
+func SaveConfigAtomic(cfg *Config, filePath string) error {
+	// Create temp file in same directory as target file
+	dir := filepath.Dir(filePath)
+	tempFile, err := os.CreateTemp(dir, filepath.Base(filePath)+".tmp.*")
```
tylerharter, line 363: rename is frequently used to atomically update a file, but I'm not sure it will work correctly if we're renaming across different file systems.  I'm not sure the temp file will always be on the same FS as the final final.  Perhaps we should write to filePath + ".tmp" so that the write goes to the same dir/FS.

src/common/config.go
```diff
@@ -305,6 +355,64 @@ func SandboxConfJson() string {
...
+	
+	// Close temp file
+	if err := tempFile.Close(); err != nil {
+		return fmt.Errorf("failed to close temp file: %v", err)
+	}
+	tempFile = nil // Mark as closed to avoid double-close in defer
+	
+	// Atomic rename - this is the key operation that prevents corruption
+	if err := os.Rename(tempPath, filePath); err != nil {
+		return fmt.Errorf("failed to rename temp file: %v", err)
```
tylerharter, line 402: %w

tylerharter: and all other places too

src/common/config.go
```diff
@@ -305,6 +355,64 @@ func SandboxConfJson() string {
...
+		return fmt.Errorf("failed to write temp file: %v", err)
+	}
+	
+	// Sync to ensure data is written to disk
+	if err := tempFile.Sync(); err != nil {
+		return fmt.Errorf("failed to sync temp file: %v", err)
+	}
+	
+	// Close temp file
+	if err := tempFile.Close(); err != nil {
```
tylerharter, line 395: This is a little complicated with Close in two places.  But I understand why you're doing it: because you want Close to happen before the rename (not when this function returns).  Maybe we should move the file write logic out to a new file, so that a defer Close() that runs at the end is appropriate.

PR #331, opened 2025-08-16
Run Openlambda as regular user

adajani:
OpenLambda was running all container processes as root, creating significant security vulnerabilities and bypassing the quota limits. You need to create 'testuser' at the host side before compiling and running the Openlambda.

Signed-off-by: Ahmed Dajani (adajain@iastate.edu)

### DIFF
--- a/examples/user/f.py
+++ b/examples/user/f.py
@@ -1,0 +1,103 @@
+import subprocess, uuid, errno, os
+
+"""
+This test is designed to test disk space exhaustion (ENOSPC) handling.
+The worker must be installed on quota enabled disk with testuser
+"""
+def f(event):
+    """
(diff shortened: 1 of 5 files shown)


kaimast:
Could you adjust the CI to run as a regular user to see if it works as intended?

tylerharter:
Hi Ahmed, thanks for the PR!  I would be pretty interested in being able to run without root, but I'm concerned about not using unshare (because then we wouldn't have new namespaces).  I dropped you an email.  Maybe we can find a time to chat about this via zoom.

tylerharter:
Actually, your email bounced back, perhaps you can email me at tharter@wisc.edu.

PR #335, opened 2025-09-09
feat: Implement full WSGI support for Python runtime

pmehrotra3:
This change makes the OL worker WSGI-compliant by:\n- Modifying the Go worker to allow flexible URL paths.\n- Upgrading the Python server with middleware to correctly handle SCRIPT_NAME and PATH_INFO for frameworks like Flask and Django.\n- Resolving dependency loading and process deadlocks.

### DIFF
--- a/go/worker/event/lambdaServer.go
+++ b/go/worker/event/lambdaServer.go
@@ -6,8 +6,8 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/open-lambda/open-lambda/go/common"
-	"github.com/open-lambda/open-lambda/go/worker/lambda"
+	"github.com/open-lambda/open-lambda/ol/common"
+	"github.com/open-lambda/open-lambda/ol/worker/lambda"
 )
 
 // LambdaServer is a worker server that listens to run lambda requests and forward
@@ -57,13 +57,12 @@ func (s *LambdaServer) RunLambda(w http.ResponseWriter, r *http.Request) {
 		// components represent run[0]/<name_of_sandbox>[1]/<extra_things>...
 		// ergo we want [1] for name of sandbox
 		urlParts := getURLComponents(r)
-		if len(urlParts) == 2 {
-			img := urlParts[1]
-			s.lambdaMgr.Get(img).Invoke(w, r)
-		} else {
-			w.WriteHeader(http.StatusInternalServerError)
-			w.Write([]byte("expected invocation format: /run/<lambda-name>"))
-		}
+		
+		// The lambda name is always the second part. The people inside (server.py)
+        // will handle the rest of the path.
+	    lambdaName := urlParts[1]
+        s.lambdaMgr.Get(lambdaName).Invoke(w, r)
+
 	}
 }
 
--- a/min-image/runtimes/python/server.py
+++ b/min-image/runtimes/python/server.py
@@ -2,7 +2,8 @@
 
 ''' Python runtime for sock '''
 
-import os, sys, json, argparse, importlib, traceback, time, fcntl, array, socket, struct
+# --- FIX 1: ADD 're' IMPORT ---
+import os, sys, json, argparse, importlib, traceback, time, fcntl, array, socket, struct, re
 
 sys.path.append("/usr/local/lib/python3.10/dist-packages")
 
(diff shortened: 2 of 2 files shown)


go/worker/event/lambdaServer.go
```diff
@@ -57,13 +57,12 @@ func (s *LambdaServer) RunLambda(w http.ResponseWriter, r *http.Request) {
...
-		if len(urlParts) == 2 {
-			img := urlParts[1]
-			s.lambdaMgr.Get(img).Invoke(w, r)
-		} else {
-			w.WriteHeader(http.StatusInternalServerError)
-			w.Write([]byte("expected invocation format: /run/<lambda-name>"))
-		}
+		
+		// The lambda name is always the second part. The code inside (server.py)
+                // will handle the rest of the path (which is possibly a Flask or Django request).
```
tylerharter, line 62: formatting

min-image/runtimes/python/server.py
```diff
@@ -14,13 +14,65 @@
...
+            m = self._any_re.match(path)
+            if not m:
+                start_response("404 Not Found", [("Content-Type", "text/plain")])
+                return [b"Not Found"]
+            prefix = m.group(0)
+            rest = path[len(prefix):] or "/"
+
+        environ["SCRIPT_NAME"] = script + prefix
+        environ["PATH_INFO"] = rest
+        environ.setdefault("HTTP_X_FORWARDWARDED_PREFIX", prefix)
```
tylerharter, line 56: what is this?

min-image/runtimes/python/server.py
```diff
@@ -14,13 +14,65 @@
 
 import ol
 
+# This is a WSGI Middleware. Think of it as a "smart translator" that sits
```
tylerharter, line 17: should we link to any documentation about WSGI?

min-image/runtimes/python/server.py
```diff
@@ -14,13 +14,65 @@
...
 
+
 def web_server():
     print(f"server.py: start web server on fd: {file_sock.fileno()}")
     sys.path.append('/handler')
+    
+    
+    # The user's dependencies (like Flask and Werkzeug) were installed into an 'app/'
+    # folder. We add that folder to Python's search path so it can find them.
+    sys.path.append('/handler/app')
```
tylerharter, line 75: I'm confused about this.  It's not under /packages?  https://github.com/open-lambda/open-lambda/blob/main/go/worker/sandbox/sockPool.go#L130

min-image/runtimes/python/server.py
```diff
@@ -14,13 +14,65 @@
...
+# to fix the URL path so your application's routing works correctly, even
+# when it's not running at the root of a domain (e.g., it's at /run/myapp).
+class PrefixWSGI:
+    
+    _any_re = re.compile(r"^/run/[^/]+")
+
+    def __init__(self, app, app_name=None):
+        self.app = app
+        self.app_name = app_name
+        self._fixed_prefix = f"/run/{app_name}" if app_name else None
```
tylerharter, line 28: Why wouldn't it have an app name always?

Or, why do we need to know the app name up front?  Couldn't we just strip out the first two parts ("run" and app_name) each time?

min-image/runtimes/python/server.py
```diff
@@ -67,16 +119,55 @@ def options(self):
...
-        app = tornado.wsgi.WSGIContainer(f.app)
+    
+        # --- This is the "Smart Path" for Flask/Django Apps ---
+        # If the "app" variable exists, we know we're dealing with a complex
+        # WSGI application that needs special handling for its URL routing.
+
+        # 1. (Optional) Read an environment variable for a "strict mode."
+        #   This allows a user to lock the middleware to one specific app name.
+        
+        app_name = os.environ.get("OL_APP_NAME")
```
tylerharter, line 137: where would this environment variable get set?

min-image/runtimes/python/server.py
```diff
@@ -67,16 +119,55 @@ def options(self):
...
+    
+    # First, we check if the user's code file ('f.py') contains a variable
+    # named "app". By convention, this is how Flask and other web frameworks
+    # expose their main application object. 
+    
     if hasattr(f, "app"):
-        # use WSGI entry
-        app = tornado.wsgi.WSGIContainer(f.app)
+    
+        # --- This is the "Smart Path" for Flask/Django Apps ---
```
tylerharter, line 130: The commenting in this section is pretty heavy.  I think an actual example would be more helpful than all the details.

min-image/runtimes/python/server.py
```diff
@@ -67,16 +119,55 @@ def options(self):
...
+        
         app = tornado.web.Application([
             (".*", SockFileHandler),
         ])
+
     server = tornado.httpserver.HTTPServer(app)
     server.add_socket(file_sock)
+    
+    
+    # The parent Go process that launched this sandbox is currently blocked,
```
tylerharter, line 163: I'm pretty confused about this part.  Why is it necessary now but not before?  If you're signaling readiness in a new way, why didn't you need to make changes on the Go side of things?

tylerharter (commented):
Good work!

go/worker/event/lambdaServer.go
```diff
@@ -6,8 +6,8 @@ import (
 	"net/http"
 	"strings"
 
-	"github.com/open-lambda/open-lambda/go/common"
-	"github.com/open-lambda/open-lambda/go/worker/lambda"
+	"github.com/open-lambda/open-lambda/ol/common"
```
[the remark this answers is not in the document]
pmehrotra3: Have made the appropriate changes

go/worker/event/lambdaServer.go
pmehrotra3: Have made the appropriate changes.

min-image/runtimes/python/server.py
pmehrotra3: We can do so if you would prefer that. The link would be : https://peps.python.org/pep-3333/

min-image/runtimes/python/server.py
pmehrotra3: Yes, we could. I think that is a better solution. I should just keep self.app_name and remove the other two lines it seems

min-image/runtimes/python/server.py
pmehrotra3: I should probably remove the if_else block because I am thinking of removing the app_name.  This code is basically trying to modify the environ to correctly set the SCRIPT_NAME and PATH_INFO variables as required by the WSGI specification. This fixes the URL routing for frameworks like Flask. I'll refactor it to use the simpler, more robust regular expression approach for all cases.

PR #338, opened 2025-09-18
fix: update fmt.Errorf/Printf usage to resolve format string errors

Yashwanth-Ranjan-Singaravel:
Updated usage of `fmt.Errorf` and `fmt.Printf` in the `go/` directory to resolve compilation issues.

### DIFF
--- a/go/boss/cloudvm/azure.go
+++ b/go/boss/cloudvm/azure.go
@@ -100,7 +100,7 @@ func Download() {
 
 func Delete() {
 	// Delete the blob
-	fmt.Printf("Deleting the blob " + blobName + "\n")
+	fmt.Printf("Deleting the blob %s\n", blobName)
 
 	_, err = blobClient.Delete(ctx, nil)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


go/boss/cloudvm/azure.go
```diff
@@ -109,7 +109,7 @@ func Delete() {
 	}
 
 	// Delete the container
-	fmt.Printf("Deleting the blob " + containerName + "\n")
+	fmt.Printf("Deleting the blob %s\n", containerName)
```
tylerharter, line 112: Just noticed, is it supposed to say "blob" again, or probably "container"?

go/worker/sandbox/dockerutil/dockerutil.go
```diff
@@ -30,7 +30,7 @@ func ImageExists(client *docker.Client, name string) (bool, error) {
 func SafeKill(client *docker.Client, cid string) error {
 	container_insp, err := client.InspectContainer(cid)
 	if err != nil {
-		return fmt.Errorf("failed to get inspect docker container ID %v: ", cid, err)
+		return fmt.Errorf("failed to get inspect docker container ID %v: %v", cid, err)
```
tylerharter, line 33: For Errorf, %w is better than %v, because then you can get a stack of errors that you could inspect at a higher level if you wanted to.

Most of the code uses %v because I didn't know about %w (or maybe it's newer?).  But we should use %w for this going forward.

tylerharter (changes requested):
Thanks!

PR #341, opened 2025-09-29
feat/fake-root-worker

guptadevagya:
adds a rootless mode for the worker (in --detach). we start the worker in a user namespace (fake root) and move the cgroup pool under the systemd user slice. if controllers aren’t delegated, we skip cgroup writes instead of crashing. includes small preflight hints and a retry if a cg-* dir already exists.


go/worker/cgroup_rootless.go
guptadevagya, line 1: rootless helpers: find user.slice, pick a cgroup pool path, and enable controllers best-effort

m0mosenpai:
One thing that I noticed while testing is that cgroup delegation under user slices in systemd, is not something that's enabled on all systems. In my case (Arch Linux), it was off by default. When that happens, `/sys/fs/cgroup/user.slice/user-%d.slice/user@%d.service` directory doesn't exist. It requires creating and configuring a conf file with root permissions and rebooting the machine. In such a case, the fallback code would be triggered and the user would either have to run OL with root or set up namespace delegation first.

PR #342, opened 2025-09-30
Added Kafka worker-side consumer logic

Yashwanth-Ranjan-Singaravel:
Added logic to run Kafka consumers during startup of workers that can act as triggers to invoke lambda functions on new messages from configured brokers.


Yashwanth-Ranjan-Singaravel:
I'll definitely clean up the code by removing redundant comments and logging statements before I mark this as ready!

go/worker/event/kafkaServer.go
```diff
@@ -0,0 +1,311 @@
...
+		if strings.HasPrefix(consumerName, lambdaName+"-") {
+			consumer.cleanup()
+			delete(ks.lambdaConsumers, consumerName)
+			slog.Info("Unregistered Kafka consumer for lambda", "lambda", lambdaName)
+		}
+	}
+}
+
+// StartConsuming starts the Kafka server (all lambda consumers are managed separately)
+func (ks *KafkaServer) StartConsuming() error {
```
tylerharter, line 283: If we never return an error, let's not have a return value

go/worker/lambda/lambdaManager.go
```diff
@@ -15,6 +15,13 @@ import (
 	"github.com/open-lambda/open-lambda/go/worker/sandbox"
 )
 
+// KafkaServerInterface defines the interface for registering Kafka triggers
```
tylerharter, line 18: If we eliminate the dependency of lambda on event, perhaps we can simplify.

go/worker/event/server.go
```diff
@@ -249,6 +250,69 @@ func RegistryHandler(w http.ResponseWriter, r *http.Request) {
...
+// and registers them immediately to avoid lazy loading delays
+func preloadAllKafkaLambdas(kafkaServer *KafkaServer, lambdaMgr *lambda.LambdaMgr) error {
+	if lambdaStore == nil {
+		return fmt.Errorf("lambda store not initialized")
+	}
+
+	slog.Info("Starting preload of Kafka lambdas from registry")
+
+	// Get all lambda names from the registry
+	lambdaNames := lambdaStore.ListAllLambdas()
```
tylerharter, line 263: I don't think this will work quite right when we run with a boss (the main use case).  Because right now, I think every worker would start doing the same working, consuming the same messages.  I think we need to expose an interface for the boss to tell us to start consuming something.

go/worker/event/kafkaServer.go
```diff
@@ -0,0 +1,311 @@
...
+	"github.com/open-lambda/open-lambda/go/worker/lambda"
+)
+
+
+// LambdaKafkaConsumer manages Kafka consumption for a specific lambda function
+type LambdaKafkaConsumer struct {
+	consumerName  string // Unique name for this consumer (may include index suffix)
+	lambdaName    string // Actual lambda function name for invocation
+	kafkaTrigger  *common.KafkaTrigger
+	client        *kgo.Client
```
tylerharter, line 25: I think we only use a couple methods on this, like PollFetches and Close.  I wonder if we should have an interface with those two methods and note in the comments that the kgo.Client implements the interface?  I'm just thinking about how we could do some unit testing, where we mock out the Client for testing purposes.

go/worker/event/kafkaServer.go
```diff
@@ -0,0 +1,311 @@
...
+	lambdaManager *lambda.LambdaMgr // Reference to lambda manager for direct calls
+	stopChan      chan struct{}
```
tylerharter, line 27: Let's add a comment about how this will be used.

go/worker/event/kafkaServer.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// KafkaServer manages multiple lambda-specific Kafka consumers
+type KafkaServer struct {
+	lambdaConsumers map[string]*LambdaKafkaConsumer // lambdaName -> consumer
+	lambdaManager   *lambda.LambdaMgr          // Reference to lambda manager
+	mu              sync.RWMutex
```
tylerharter, line 34: Is this just for protecting lambdaConsumers?  Need commenting about what a lock protects.  Let's just use a simple mutex too for simplicity (not RW).

go/worker/event/kafkaServer.go
```diff
@@ -0,0 +1,311 @@
...
+		lambdaManager:   lambdaManager,
+		stopChan:        make(chan struct{}),
+	}
+
+	slog.Info("Kafka server initialized")
+	return server, nil
+}
+
+// StartConsuming starts consuming messages for this lambda's Kafka triggers
+func (lkc *LambdaKafkaConsumer) StartConsuming() error {
```
tylerharter, line 93: If we can never return an error, better not to have a return value.

tylerharter (changes requested):
Great work!

go/common/lambdaConfig.go
```diff
@@ -142,12 +142,18 @@ func ExtractConfigFromTarGz(tarPath string) (*LambdaConfig, error) {
 			return nil, fmt.Errorf("invalid tar: %w", err)
 		}
 
-		if header.Name == LambdaConfigFilename {
+		// Debug: Log all files in the tarball
+		slog.Info("DEBUG: Found file in tarball", "filename", header.Name, "tarPath", tarPath)
```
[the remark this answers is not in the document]
Yashwanth-Ranjan-Singaravel: It was just a temporary debug statement to figure out where the issue I was facing was originating from. I think we can remove the statement entirely now!

go/worker/event/server.go
Yashwanth-Ranjan-Singaravel: Yeah, preloading the consumers does not make sense. I'll implement an API to register consumers instead!

tylerharter (changes requested):
Great work!

PR #343, opened 2025-10-04
Added support for using UDS instead of TCP for boss/worker operations

bridgetwu33:
Modified worker and boss config structs to take in a socket option for a filesystem path to a sock file. (to prevent admin ports from being exposed over TCP). Also modified boss to serve on UDS ports instead of HTTP. The /pid endpoint in worker now listens on a UDS sock instead of port, and the other HTTP endpoints are still on TCP as of now.


go/worker/event/server.go
```diff
@@ -299,28 +300,98 @@ func Main() (err error) {
...
+	signalChannel := make(chan os.Signal, 1)
+	signal.Notify(signalChannel, os.Interrupt, syscall.SIGTERM, syscall.SIGINT)
 
-	// if ListenAndServer returned, there must have been some issue
-	// (probably a port collision)
+	// start serving on the UNIX domain socket server
+	go func() {
+		slog.Info("worker listening on UNIX domain socket", "socket", sockPath)
+		err := udsServer.Serve(ln)
+		if err != nil && err != http.ErrServerClosed {
```
tylerharter: When we check if != nil like this, it feels to the reader that nil is a valid return value.  But the documentation says it is not.  We should quote the documentation here, and either panic if we get nil (since it violates our assumptions), or send an error saying something like "Server unexpectedly returned nil".

go/worker/event/server.go
```diff
@@ -199,20 +198,11 @@ func shutdown(pidPath string, server cleanable) {
...
 
-	slog.Info(fmt.Sprintf("Remove %s.", pidPath))
-	if err := os.Remove(pidPath); err != nil {
-		slog.Error(fmt.Sprintf("error: %s", err))
-		rc = 1
-	}
-
-	slog.Info(fmt.Sprintf("Exiting worker (PID %d)", os.Getpid()))
-	os.Exit(rc)
+	slog.Info(fmt.Sprintf("Printed final stats of worker (PID %d)", os.Getpid()))
```
tylerharter: A lot of current code uses this format (slog.Info(fmt.Sprintf(...)), but that's just because I was porting from log.Printf.

A better pattern is like this:
slog.Info("Printed final stats of worker", "pid", os.Getpid())

PR #345, opened 2025-10-08
added systemd user cgroup delegation support for rootless mode

m0mosenpai:


go/common/cgroups_rootless.go
```diff
@@ -0,0 +1,44 @@
...
+	}
+	return os.Getuid()
+}
+
+// Returns the systemd user slice path for this user.
+func CgroupPath() string {
+	uid := HostUID()
+	base := fmt.Sprintf("/sys/fs/cgroup/user.slice/user-%d.slice/user@%d.service", uid, uid)
+	if st, err := os.Stat(base); err != nil || !st.IsDir(){
+		panic(fmt.Errorf("%s: no delegated user slice at uid %d", err, uid))
```
m0mosenpai, line 41: cgroup delegation in systemd is not enabled by default in some systems. The user will have to use root privileges to configure that on their own before these user slices are available for use. If we plan to drop support for running OL as root completely, then this would result in an error.

go/worker/sandbox/cgroups/pool.go
```diff
@@ -41,13 +40,11 @@ func NewCgroupPool(name string) (*CgroupPool, error) {
 	if err := syscall.Mkdir(groupPath, 0700); err != nil {
 		return nil, fmt.Errorf("Mkdir %s: %s", groupPath, err)
 	}
-
 	// Make controllers available to child groups
 	rpath := fmt.Sprintf("%s/cgroup.subtree_control", groupPath)
-	if err := ioutil.WriteFile(rpath, []byte("+pids +io +memory +cpu"), os.ModeAppend); err != nil {
-		panic(fmt.Sprintf("Error writing to %s: %v", rpath, err))
+	if err := os.WriteFile(rpath, []byte("+pids +io +memory +cpu"), os.ModeAppend); err != nil {
```
m0mosenpai, line 45: There is also a scenario where a user might have systemd delegation configured but they don't have the appropriate controllers made available. Right now this would cause OL to panic here. One way around this would be to first try and probe the system to see which controllers are available and then try to write to only those. If none are available, we fail.

go/worker/commands.go
```diff
@@ -113,17 +123,33 @@ func upCmd(ctx *cli.Context) error {
...
+		// legacy used CLONE_NEWNS so mounts don't spam systemd.
+		// now, if --rootless, also create a user+UTS namespace and map uid/gid -> 0 (fake root).
 		attr := os.ProcAttr{
 			Files: []*os.File{nil, f, f},
-			Sys: &syscall.SysProcAttr{
-				Unshareflags: syscall.CLONE_NEWNS,
-			},
 		}
+		if rootless {
+			attr.Sys = &syscall.SysProcAttr{
```
tylerharter: This code is the crux of what you're doing.  I think it needs some more commenting.  E.g., will it be obvious to people that "0" is the ID for root?  Also need to explain what a user namespace is, and why that lets us run rootless.

go/common/cgroups_rootless.go
```diff
@@ -0,0 +1,31 @@
...
+			}
+		}
+	}
+
+	// Use default path if run as root
+	if os.Getuid() == 0 {
+		return "/sys/fs/cgroup"
+	}
+
+	panic(fmt.Errorf("systemd user cgroup delegation not available - cannot run rootless"))
```
tylerharter: I think instead of trying to integrate with systemd directly, I think it is more flexible if we let users configure the root cgroup to use.  We can provide instructions for users to run "systemctl show --user --property=ControlGroup" and find this location and configure their OL accordingly.

go/worker/helpers.go
```diff
@@ -40,22 +39,22 @@ func initOLBaseDir(baseDir string, dockerBaseImage string) error {
 
 	// PART 2: various files/dirs on top of the extracted image
 	fmt.Printf("\tCreate handler/host/packages/resolve.conf over base image.\n")
-	if err := os.Mkdir(path.Join(baseDir, "handler"), 0700); err != nil {
+	if err := os.Mkdir(path.Join(baseDir, "handler"), 0755); err != nil {
```
tylerharter: Why do we need to change all these permissions?  If they do everything (init and up) as root, it is fine, right?

And if they do init with sudo, and up with ID mapping, wouldn't everything be created by user 0 and used by user 0?  Not sure we want to make all these things visible to every user running on Linux.

go/worker/helpers.go
```diff
@@ -162,6 +161,22 @@ func initOLDir(olPath string, dockerBaseImage string, newBase bool) (err error)
 		fmt.Printf("\tReusing prior base at %s (pass -b to reconstruct this)\n", baseDir)
 	}
 
+	// For rootless mode: make lambda base readable so worker can bind mount it
+	if err := exec.Command("chmod", "-R", "755", baseDir).Run(); err != nil {
+		return fmt.Errorf("failed to chmod base dir: %v", err)
+	}
+
+	// For rootless mode: if run with sudo, give olPath and worker dir to actual user
```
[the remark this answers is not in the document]
m0mosenpai: @tylerharter I fixed the permissions above (reverted them back to 700s so the files are not visible to everyone else). But the base directory is created with 700 permission and since we sudo ./ol worker init, ./ol worker up doesn't work without root.

This changes was meant to give the ownership to the real user so ./ol worker up can work without issues. I'm not sure if there's a better way to get around this.

PR #346, opened 2025-10-08
types: introduce common.Limits (no behavior change)

guptadevagya:


go/common/config.go
```diff
@@ -113,26 +114,42 @@ type StorageConfig struct {
...
+	// process & memory / CPU controls
+	Procs       int `json:"procs" yaml:"procs"`
+	Mem_mb      int `json:"mem_mb" yaml:"mem_mb"`
+	CPU_percent int `json:"cpu_percent" yaml:"cpu_percent"`
+	Swappiness  int `json:"swappiness" yaml:"swappiness"`
 
-	// what percent of a core can it use per period?  (0-100, or more for multiple cores)
-	CPU_percent int `json:"cpu_percent"`
+	// worker default for runtime cap (seconds). Per-lambda may set runtime_sec.
+	Max_runtime_default int `json:"max_runtime_default" yaml:"max_runtime_default"`
```
tylerharter: Do we need a default runtime?  I thought we would have a different struct instance for defaults, not a different field within the struct?

Is there a reason we would have a regular and default for runtime, but not for other fields like procs?

go/common/config.go
```diff
@@ -279,6 +326,22 @@ func LoadGlobalConfig(path string) error {
 		return err
 	}
 
+	// Ensure mem_pool_mb meets the minimum based on current limits.
```
tylerharter: This is pretty confusing to me, and doesn't seem like it belongs here.  How do you know the machine has enough RAM to bump it up?  Even if they do, if they set a low mem pool size, they probably are reserving memory for some other system being deployed alongside OpenLambda.  We should inform the user (as checkConf does), but don't second guess them.

go/common/config.go
```diff
@@ -322,10 +385,13 @@ func checkConf(cfg *Config) error {
 		// otherwise anything running will immediately be
 		// evicted.
 		//
-		// TODO: revise evictor and relax this
-		minMem := 2 * Max(cfg.Limits.Installer_mem_mb, cfg.Limits.Mem_mb)
+		// We check against both the regular user limits and the installer limits.
```
tylerharter: Why are we getting rid of the TODO?

PR #411, opened 2026-02-06
Added lambda that integrates Forklift

JaimeZepeda08:


examples/forklift-service/f.py
```diff
@@ -0,0 +1,313 @@
...
+    return root.to_dict()
+
+
+def f(event):
+    """
+    Lambda function entry point for generating Forklift zygote trees.
+    
+    Expected event format:
+    {
+        "workload": { ... workload data ... },
```
tylerharter: Can you show the next level of detail (below) about how workload data is structured?

examples/forklift-service/f.py
```diff
@@ -0,0 +1,313 @@
...
+    
+    Returns:
+        {
+            "status": "success" | "error",
+            "result": { ... tree structure ... },
+            "error": "..." (if error)
+        }
+    """
+    try:
+        if isinstance(event, str):
```
tylerharter: Why do we need to check?  Are there different possible inputs?

examples/forklift-service/f.py
```diff
@@ -0,0 +1,313 @@
...
+            return {"status": "error", "error": "Missing 'num_nodes' in event data"}
+        
+        if not isinstance(num_nodes, int) or num_nodes < 1:
+            return {"status": "error", "error": "'num_nodes' must be a positive integer"}
+        
+        multi_package = event.get("multi_package", True)
+        
+        result = generate_tree(workload_data, num_nodes, multi_package)
+        
+        return {
```
tylerharter: Can we write a flask lambda instead?  Then, we could return a status code, like 500 or something.  The benefit is that somebody could do a "curl ... > tree.json" to get a tree, without needing to extract the result separately.

examples/forklift-service/f.py
```diff
@@ -0,0 +1,313 @@
...
+            "status": "error",
+            "error": str(e),
+            "traceback": traceback.format_exc()
+        }
+
+
+# Testing code
+if __name__ == "__main__":
+    with open("wl.json", "r") as file:
```
tylerharter: why not sys.argv[1]?

examples/forklift-service/f.py
```diff
@@ -0,0 +1,313 @@
+import json
+import heapq
+import os
+from collections import defaultdict
+
```
tylerharter: let's link to the paper up front

examples/forklift-service/f.py
```diff
@@ -0,0 +1,313 @@
...
+        for version, dep_strings in versions.items():
+            deps[pkg_name][version] = []
+            for dep_str in dep_strings.keys():
+                dep_packages = set(dep_str.split(",")) if dep_str else set()
+                deps[pkg_name][version].append(dep_packages)
+    
+    return deps
+
+
+def generate_tree(workload_data, num_nodes, multi_package=True):
```
tylerharter: This function doesn't do too much, and is only called once.  Perhaps just inline the logic where you currently call it?

examples/forklift-service/f.py
```diff
@@ -0,0 +1,313 @@
...
+        pkg_name = child_pkgV.split("==")[0]
+        version = child_pkgV.split("==")[1] if "==" in child_pkgV else None
+        
+        # skip if this package but different version is loaded
+        # TODO: see if we can handle multiple versions of the same package
+        if pkg_name in loaded_names:
+            continue
+        
+        # keep track of packages that would be loaded by this candidate
+        packages_to_load = set([child_pkgV])
```
tylerharter: Can create a set without converting:

{child_pkgV}

examples/forklift-service/f.py
```diff
@@ -0,0 +1,313 @@
+import json
+import heapq
+import os
+from collections import defaultdict
+
+
+# TODO: instead of loading deps from a file, use pip-compile on the fly
```
tylerharter: I'm thinking lets require callers to pass deps in.  Then, however we figure out those dependencies, it will be outside the nice, clean tree-building logic here.

tylerharter (changes requested):
Good work!  Haven't read everything super closely yet, but leaving some initial feedback.

Let's not have big data files here, like deps.json and wl.json.  In the docs, we can describe a user would use your lambda.  wget those files (from wherever they are, probably on the forklift repo), and then curl to your function.

tylerharter (changes requested):
Getting closer!

examples/forklift/requirements.txt
```diff
@@ -0,0 +1 @@
+flask
```
tylerharter, line 1: Let's do the pattern of requirements.in, and a pip compiled requirements.txt, as in other example lambdas.

examples/forklift/f.py
```diff
@@ -0,0 +1,318 @@
...
+from collections import defaultdict, namedtuple
+from flask import Flask, request, jsonify
+
+Candidate = namedtuple('Candidate', ['parent', 'child_pkgV', 'utility'])
+QueueEntry = namedtuple('QueueEntry', ['neg_utility', 'uid', 'candidate'])
+
+app = Flask(__name__)
+
+
+class CallMatrix:
```
tylerharter, line 16: I think a generic SparseMatrix would allow us to better mirror the logic in the paper.  The class has package-specific awareness, and therefore takes on some responsibilities handled by other functions in the pseudocode.

examples/forklift/f.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+class ZygoteTree: 
+    def __init__(self, calls, deps):
+        self.calls = calls
+        self.deps = deps
+        self.root = None
+        self.candidate_queue = []
+    
+    def _enqueue_top_child_candidate(self, parent):
```
tylerharter, line 110: This function is quite a bit longer and more complicated than the pseudocode for enqueue_top_child_candidate in the paper.  I think it is because this function is taking on more concerns, not just the core logic.

E.g., you have separate parse functions.  Splitting pkg name from version feels like it should be handled there, not in the core logic of this function.

examples/forklift/f.py
```diff
@@ -0,0 +1,318 @@
...
+        # build tree
+        tree = ZygoteTree(calls, deps)
+        tree.build_tree(num_nodes)
+
+        result = tree.to_dict()
+
+        return jsonify(result), 200
+
+    except Exception as e:
+        import traceback
```
tylerharter, line 314: imports should be top of file

examples/forklift/f.py
```diff
@@ -0,0 +1,318 @@
...
+            },
+            ...
+        },
+        "num_nodes": <int>
+    }
+    '''
+
+    try:
+        event = request.get_json()
+        if event is None:
```
tylerharter, line 291: we have a couple special cases where we return different errors for specific kinds of invalid data.  But given all is in a try/except, do we need special code for these?

examples/forklift/f.py
```diff
@@ -0,0 +1,318 @@
...
+                ...
+            },
+            ...
+        },
+        "num_nodes": <int>
+    }
+    '''
+
+    try:
+        event = request.get_json()
```
tylerharter, line 290: perhaps the body of this try should be it's own function to cleanly separate logic: one function for core logic, another (which calls the former) that adds error handling.

examples/forklift/f.py
tylerharter: Maybe a helper function or two would help as well?

tylerharter (changes requested):
Nice!  Getting close.

examples/forklift/f.py
```diff
@@ -0,0 +1,230 @@
...
+QueueEntry = namedtuple('QueueEntry', ['neg_utility', 'uid', 'candidate'])
+
+app = Flask(__name__)
+
+
+def parse_workload(workload):
+    func_packages = {}
+
+    for func in workload.get("funcs", []):
+        func_name = func.get("name", "")
```
tylerharter, line 22: is a function without a name valid input?  If somebody gives us a garbage file, do we want to silently carry on, or crash in an obvious way?

I'm not a fan of the .get's -- if there are specific cases where it is valid to have a field or not, then let's comment to justify it.

examples/forklift/f.py
```diff
@@ -0,0 +1,230 @@
...
+from flask import Flask, request, jsonify
+
+Candidate = namedtuple('Candidate', ['parent', 'child_pkgV', 'utility'])
+QueueEntry = namedtuple('QueueEntry', ['neg_utility', 'uid', 'candidate'])
+
+app = Flask(__name__)
+
+
+def parse_workload(workload):
+    func_packages = {}
```
tylerharter, line 19: what are the keys/values going to be?  A comment would be nice.

examples/forklift/f.py
```diff
@@ -0,0 +1,230 @@
...
+import pandas as pd
+from flask import Flask, request, jsonify
+
+Candidate = namedtuple('Candidate', ['parent', 'child_pkgV', 'utility'])
+QueueEntry = namedtuple('QueueEntry', ['neg_utility', 'uid', 'candidate'])
+
+app = Flask(__name__)
+
+
+def parse_workload(workload):
```
tylerharter, line 18: what is it going to return?  describe the pandas df, and what the rows+columns mean

examples/forklift/f.py
```diff
@@ -0,0 +1,230 @@
...
+        packages = set()
+        meta = func.get("meta", {})
+        req_txt = meta.get("requirements_txt", "")
+        
+        for line in req_txt.split("\n"):
+            line = line.strip()
+            if "==" in line and not line.startswith("#"):
+                packages.add(line)
+        
+        if packages:
```
tylerharter, line 35: do we need the conditional?  Is it a problem to have a function with no packages?

examples/forklift/f.py
```diff
@@ -0,0 +1,230 @@
...
+    # count call frequencies
+    call_counts = defaultdict(int)
+    for call in workload.get("calls", []):
+        name = call.get("name", "")
+        if name:
+            call_counts[name] += 1
+    
+    # rows=calls, columns=package==version, values=0/1
+    rows = {}
+    if call_counts:
```
tylerharter, line 47: is it valid to have an input with zero calls?  Do we want to complicate the function with this case?

examples/forklift/f.py
```diff
@@ -0,0 +1,230 @@
...
+        for func_name, packages in func_packages.items():
+            count = max(1, call_counts.get(func_name, 1))
+            for i in range(count):
+                rows[f"{func_name}_{i}"] = packages.copy()
+    else:
+        # if no call counts, just add one call per function
+        for func_name, packages in func_packages.items():
+            rows[func_name] = packages.copy()
+    
+    if not rows:
```
tylerharter, line 57: another check I'm not sure about

examples/forklift/f.py
```diff
@@ -0,0 +1,230 @@
...
+        return pd.DataFrame(dtype=int)
+    
+    # sorted list of all unique packages across all calls, to ensure consistent column ordering
+    all_pkgs = sorted(set().union(*rows.values()))
```
tylerharter, line 61: I'm working too hard to parse this, maybe a few lines will be more readable?

tylerharter (commented):
Let's iterate a bit more to make sure the code is elegant and commented (at key places).  Left a few ideas in the parsing code.

PR #425, opened 2026-03-03
 `sandboxset`: thread-safe sandbox pool (issue #217)

AmiBuch:
# `sandboxset`: thread-safe sandbox pool (issue #217)

A per-Lambda pool of sandbox *slots*. Callers ask for one, use it, hand it back. The set pauses/unpauses and recycles slots.

## Usage

```go
set := sandboxset.New(&sandboxset.Config{
    Pool:        myPool,
    CodeDir:     "/path/to/lambda",
    ScratchDirs: myScratchDirs,
})
defer set.Close()

ref, err := set.GetOrCreateUnpaused()
if err != nil { /* ... */ }

resp, err := ref.Sandbox().Client().Do(req)
if err != nil {
    ref.Sandbox().Destroy("handler failed")
    ref.MarkDead()
}
ref.Put()
```

## API

```go
type SandboxSet interface {
    GetOrCreateUnpaused() (*SandboxRef, error)
    Close() error
}

type SandboxRef struct { /* all unexported */ }

func (r *SandboxRef) Sandbox() sandbox.Sandbox  // underlying container
func (r *SandboxRef) MarkDead()                 // tell the set the sandbox is unusable
func (r *SandboxRef) Put()                      // return the ref to its set
```

### Config

| Field         | Purpose                                                       |
|---------------|---------------------------------------------------------------|
| `Pool`        | Underlying `sandbox.SandboxPool` that creates containers      |
| `CodeDir`     | Lambda handler code directory                                 |
| `ScratchDirs` | `*common.DirMaker`; creates a writable dir per new sandbox    |
| `Meta`        | Runtime config (memory, packages, etc.) — optional            |
| `Parent`      | Optional parent `SandboxSet` to fork from; must outlive child |
| `IsLeaf`      | Marks sandboxes as non-forkable                               |

## Flow

```
GetOrCreateUnpaused
   └── claimIdle          (lock; reuse idle ref, empty slot, or append new)
   └── Unpause existing   (on failure: log, null sb, fall through to create)
   └── createSandbox      (on failure: releaseSlot, return err)

Put
   ├── sb == nil  → releaseSlot       (slot clear; no Pause)
   └── sb != nil  → put
                     ├── set closed? → releaseSlotLocked (no Pause)
                     ├── Pause fails? → releaseSlot
                     └── otherwise → inUse = false (recycled)
```

## Testing

```bash
# Unit tests (MockSandboxPool)
cd go && go test ./worker/sandboxset/tests/ -v -race -count=1

# Integration tests (real DockerPool, requires ol-min image)
cd go && go test ./worker/sandboxset/tests/ -v -tags=integration -count=1
```

## Next steps

- **Step 2**: Replace `LambdaInstance` goroutines with a `SandboxSet`.
- **Step 3**: Use `SandboxSet` as the node in the zygote tree (via `Config.Parent`).


tylerharter (changes requested):

go/worker/sandboxset/api.go
```diff
@@ -0,0 +1,128 @@
...
+	//
+	// Use this when a request produced an unrecoverable error and
+	// the sandbox should not be reused. "reason" is a
+	// human-readable explanation that shows up in later error
+	// messages (same convention as sandbox.Sandbox.Destroy).
+	//
+	// If the sandbox is not in the pool it is still destroyed —
+	// resources are always freed. The returned error is
+	// informational only.
+	Destroy(sb sandbox.Sandbox, reason string) error
```
tylerharter: don't need

go/worker/sandboxset/tests/sandboxset_test.go
```diff
@@ -0,0 +1,449 @@
+package tests
```
tylerharter: let's do 2-3 tests initially

tylerharter (commented):
I think I forgot to click submit on the feedback, sorry!

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+	return &sandboxSetImpl{cfg: cfg}, nil
+}
+
+// makeScratchDir creates a scratch directory for a new sandbox.
+// DirMaker.Make panics on failure (e.g., disk full), so we recover
+// here and return an error instead of crashing the worker.
+func (s *sandboxSetImpl) makeScratchDir() (dir string, err error) {
+	defer func() {
+		if r := recover(); r != nil {
```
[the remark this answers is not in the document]
tylerharter: because we use panics to indicate there is a bug or the state is something that should never happen.

We use errors for recoverable issues.

tylerharter (commented):
Good work!  Getting closer.

tylerharter (changes requested):
Great work!

go/worker/sandboxset/api.go
```diff
@@ -0,0 +1,141 @@
+// Package sandboxset provides a thread-safe pool of sandboxes for a single
+// Lambda function.
+//
+// A SandboxSet replaces per-instance goroutines with a simple pool.
+// Callers just ask for a sandbox and don't worry about whether it is
+// freshly created or recycled from a previous request.
+//
+// Sandbox lifecycle inside a SandboxSet:
```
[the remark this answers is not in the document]
tylerharter: The usage is good because that's what API users need.  The internals in the flow diagram are not something users of your code should worry about.  Those are internal details.

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,225 @@
...
+	return r.sb
+}
+
+// Put returns the sandbox to its parent set.
+// This is a convenience method equivalent to set.Put(ref.Sandbox()).
+func (r *SandboxRef) Put() error {
+	return r.set.Put(r.sb)
+}
+
+// Destroy removes the sandbox from its parent set and destroys it.
```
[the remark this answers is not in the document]
tylerharter: I still don't think we need it.  Please justify why, or remove.

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,261 @@
...
+// sandboxSetImpl is the private concrete type returned by New.
+type sandboxSetImpl struct {
+	cfg *Config
+
+	mu     sync.Mutex // protects below fields
+	pool   []*SandboxRef
+	closed bool
+}
+
+func newSandboxSet(cfg *Config) (*sandboxSetImpl, error) {
```
tylerharter: We should return errors for things we expect to happen during normal operation.  For bugs, we actually want to just crash.  So newSandboxSet should not return an error.  You can panic if you get bad args.  Or, just let the panic happen itself later when there is a nil deref.

go/worker/sandboxset/api.go
```diff
@@ -0,0 +1,79 @@
...
+
+	// ScratchDirs creates a unique writable directory for each
+	// new sandbox. The set calls ScratchDirs.Make internally
+	// so that GetOrCreateUnpaused can remain argument-free.
+	ScratchDirs *common.DirMaker
+}
+
+// New creates a SandboxSet from cfg. Returns an error if any of
+// Pool, CodeDir, or ScratchDirs are missing.
+func New(cfg *Config) (SandboxSet, error) {
```
tylerharter: No error, should always succeed (or crash).

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,261 @@
...
+	inUse bool
+}
+
+// Sandbox returns the underlying sandbox.
+func (r *SandboxRef) Sandbox() sandbox.Sandbox {
+	return r.sb
+}
+
+// Put returns the sandbox to its parent set, or destroys it if Broken is true.
+func (r *SandboxRef) Put() error {
```
tylerharter, line 38: I like this being a very simple wrapper around the private put method in sandboxSetImpl.  The code is easier to understand if locking is a detail internal to sandboxSetImpl, not split across multiple classes.

The only purpose/logic of this function: identifying which sandboxset should do the put.

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,261 @@
...
+		return nil, err
+	}
+	return sb, nil
+}
+
+// GetOrCreateUnpaused implements SandboxSet.
+// Fast path: claim an idle ref with an existing sandbox via tryClaimIdle, then Unpause outside the lock.
+// Slow path (nil ref): the claimed ref has no sandbox (either destroyed or newly added). A new sandbox is created outside the lock and assigned to the ref.
+func (s *sandboxSetImpl) GetOrCreateUnpaused() (*SandboxRef, error) {
+	for {
```
tylerharter: Let's not do retry for now.  lambdaInstance.go already has retry logic for if Unpause fails.  When we use sandbox set at a higher layer, then we can make a coherent decision about the 1 place to do retry (in sandbox set, or in the user of sandbox set).

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,261 @@
...
+		}
+		if ref.sb != nil {
+			ref.inUse = true
+			return ref
+		}
+		if nilRef == nil {
+			nilRef = ref
+		}
+	}
+	if nilRef != nil {
```
tylerharter: The naming is confusing me here.  Why wouldn't a nilRef be nil?

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+	if cfg.ScratchDirs == nil {
+		return nil, fmt.Errorf("sandboxset: Config.ScratchDirs must not be nil")
+	}
+	return &sandboxSetImpl{cfg: cfg}, nil
+}
+
+// claimIdle returns an idle ref from the pool, marking it inUse.
+// Caller must hold s.mu.
+func (s *sandboxSetImpl) claimIdle() *SandboxRef {
```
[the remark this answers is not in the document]
tylerharter: I was imagining those 3 paths in claimIdle.  And I think claimIdle can do the locking.  We can condense claimIdle, tryClaimIdle, and appendNilRef to just this function.

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,261 @@
...
+
+		if ref == nil {
+			// No idle ref — add a new nil ref to the pool.
+			ref, err = s.appendNilRef()
+			if err != nil {
+				return nil, err
+			}
+		}
+
+		s.mu.Lock()
```
tylerharter: Do we need locking?  I was imagining the design would be that when not inUse, any thread might want to use it, so locking would be needed.  But when inUse, it would effectively be reserved by a single thread, so no more locking until returned to not inUse.

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,261 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return sb, nil
+}
+
+// GetOrCreateUnpaused implements SandboxSet.
+// Fast path: claim an idle ref with an existing sandbox via tryClaimIdle, then Unpause outside the lock.
+// Slow path (nil ref): the claimed ref has no sandbox (either destroyed or newly added). A new sandbox is created outside the lock and assigned to the ref.
+func (s *sandboxSetImpl) GetOrCreateUnpaused() (*SandboxRef, error) {
```
tylerharter, line 129: Really, the logic should break into two steps:
1. get a SandboxRef (with our without a Sandbox)
2. make sure that SandboxRef has a healthy, unpaused Sandbox inside it

Step 1 requires locking, step 2 does not.

GetOrCreateUnpaused could be very short, calling functions corresponding to the two steps.  Or, GetOrCreateUnpaused could call step 1, then do step 2 inline.

tylerharter (changes requested):
API is looking great!  Lets iterate on the implementation a bit more.

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	scratchDir := s.cfg.ScratchDirs.Make("sb")
+	sb, err := s.cfg.Pool.Create(
+		parentSb, s.cfg.IsLeaf,
+		s.cfg.CodeDir, scratchDir,
+		s.cfg.Meta,
+	)
+	if err != nil {
+		return nil, err
```
tylerharter: wrap err with %w

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,218 @@
...
+	ref, err := s.claimIdle()
+	if err != nil {
+		return nil, err
+	}
+
+	// Step 2: make sure the ref has a healthy, unpaused Sandbox.
+	if ref.sb != nil {
+		if err := ref.sb.Unpause(); err != nil {
+			_ = s.destroy(ref, fmt.Sprintf("unpause: %v", err))
+			return nil, fmt.Errorf("sandboxset: unpause: %w", err)
```
tylerharter: Why return an error, instead of just continuing on to make a good sandbox?

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,218 @@
...
+		}
+		return ref, nil
+	}
+
+	newSb, err := s.createSandbox()
+	if err != nil {
```
tylerharter, line 147: Why not use our existing logic?  Set Broken=true and put the sandbox ref.

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	ref.sb = newSb
+	return ref, nil
+}
+
+// put pauses the sandbox and returns the ref to the idle pool.
+// Pause is called before re-acquiring the lock to avoid blocking pool access during I/O.
+func (s *sandboxSetImpl) put(ref *SandboxRef) error {
+	s.mu.Lock()
```
tylerharter: do we need locking?  I thought this can only be called if inUse is true, and if inUse is true, isn't there just one thread using it?

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,218 @@
...
+	if err := sb.Pause(); err != nil {
+		_ = s.destroy(ref, fmt.Sprintf("pause failed: %v", err))
+		return fmt.Errorf("sandboxset: sandbox destroyed because Pause failed: %w", err)
+	}
+
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.closed {
+		return fmt.Errorf("sandboxset: closed (sandbox destroyed by Close)")
```
tylerharter: should we be destroying the sandbox or something?

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,225 @@
...
+		return fmt.Errorf("sandboxset: sandbox %s not found in pool (still destroyed)", sb.ID())
+	}
+	return nil
+}
+
+// Close implements SandboxSet.
+//
+// All sandboxes are snapshot under the lock, then destroyed outside it.
+func (s *sandboxSetImpl) Close() error {
+	s.mu.Lock()
```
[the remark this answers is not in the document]
tylerharter: Please comment on this before resolving.

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,218 @@
...
+}
+
+// Close implements SandboxSet.
+func (s *sandboxSetImpl) Close() error {
+	s.mu.Lock()
+	if s.closed {
+		s.mu.Unlock()
+		return fmt.Errorf("sandboxset: already closed")
+	}
+	s.closed = true
```
tylerharter, line 210: wouldn't this need to be lock protected?

tylerharter (changes requested):

go/worker/sandboxset/sandboxset.go
AmiBuch: Got it.

go/worker/sandboxset/sandboxset.go
AmiBuch: Changed. On unpause failure, now it goes to createSandbox path. Mirrors the LambdaInstance way of unpause failure.

go/worker/sandboxset/sandboxset.go
AmiBuch: Got it.

go/worker/sandboxset/sandboxset.go
AmiBuch: Changed lock to only affect s.closed and ref.inUse

go/worker/sandboxset/sandboxset.go
AmiBuch: I couldn't reuse s.destroy(ref, ...) here because that helper clears inUse=false, which would let another goroutine grab the ref mid-recovery. So I null out ref.sb under the lock and then call Destroy outside the lock, keeping inUse=true throughout.

go/worker/sandboxset/sandboxset.go
AmiBuch: I have switched to ref.Broken = true; ref.Put() so there's only one cleanup path.

go/worker/sandboxset/sandboxset.go
AmiBuch: Chose this flow : null out ref.sb, release the lock, and call sb.Destroy("sandboxset closed during put")

go/worker/sandboxset/sandboxset.go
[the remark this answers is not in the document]
AmiBuch: Close only runs at shutdown so there's no concurrent workload to protect. Simplified Close() to hold s.mu across the whole body with a defer, one loop, no more toDestroy slice.

go/worker/sandboxset/sandboxset.go
AmiBuch: Close now holds the lock across the whole function

go/worker/sandboxset/sandboxset.go
```diff
@@ -0,0 +1,183 @@
...
+		ref.sb = newSb
+	}
+
+	return ref, nil
+}
+
+func (s *sandboxSetImpl) put(ref *SandboxRef) {
+	if ref.sb != nil {
+		if err := ref.sb.Pause(); err != nil {
+			ref.sb.Destroy("sandboxset: pause failed")
```
tylerharter: if Pause fails, it should already be destroyed automatically, right?

tylerharter (approved):
Great work!

PR #440, opened 2026-04-09
A simple kafka lambda that interacts with a DB

Yashwanth-Ranjan-Singaravel:
A simple lambda that maintains a running tally of numbers using a local Postgress DB


examples/kafka-db-sum/f.py
```diff
@@ -0,0 +1,198 @@
...
+    return psycopg2.connect(DATABASE_URL)
+
+
+def ensure_db():
+    """Create the running_sum table if it doesn't exist (runs once per sandbox)."""
+    global _db_initialized
+    if _db_initialized:
+        return
+    conn = get_db()
+    try:
```
tylerharter, line 26: I don't love try/finally.  Does "with" work on conn?  I.e., is conn a "context manager" that can close automatically with used with "with"?

examples/kafka-db-sum/f.py
```diff
@@ -0,0 +1,198 @@
...
+
+    # POST — process a Kafka message containing a number
+    offset = int(request.headers.get("X-Kafka-Offset", "-1"))
+    topic = request.headers.get("X-Kafka-Topic", "unknown")
+    partition = request.headers.get("X-Kafka-Partition", "unknown")
+
+    body = request.get_json(silent=True)
+
+    # Accept {"number": N}
+    if isinstance(body, dict):
```
tylerharter, line 105: why are we doing these checks?  Is any other content valid?

examples/kafka-db-sum/f.py
```diff
@@ -0,0 +1,198 @@
...
+    finally:
+        conn.close()
+
+
+@app.route("/", methods=["GET", "POST"])
+def handle():
+    ensure_db()
+
+    # GET — return current state (useful for checking progress via HTTP)
+    if request.method == "GET":
```
tylerharter, line 74: why not just make a separate endpoint, and keep this one clean?

examples/kafka-db-sum/f.py
```diff
@@ -0,0 +1,198 @@
...
+    conn = None
+    try:
+        conn = get_db()
+        with conn.cursor() as cur:
+            # Lock the row and read last processed offset
+            cur.execute(
+                "SELECT last_offset FROM running_sum WHERE id = 1 FOR UPDATE"
+            )
+            row = cur.fetchone()
+            last_offset = row[0] if row else -1
```
tylerharter, line 117: do we need to check row?  does fetchone() raise an exception or just return None when there is no row?

tylerharter (approved):
Perhaps you can cleanup in a following PR

PR #441, opened 2026-04-09
Added worker specific kafka config

Yashwanth-Ranjan-Singaravel:
This adds the following config options and defaults to the worker side Kafka config:
  - cache_enabled: true
  - cache_size: 1024
  - session_timeout_sec: 10
  - heartbeat_interval_sec: 3
  - poll_timeout_sec: 1

### DIFF
--- a/go/worker/event/kafkaServer.go
+++ b/go/worker/event/kafkaServer.go
@@ -89,13 +89,15 @@ func (km *KafkaManager) newLambdaKafkaConsumer(consumerName string, lambdaName s
 		return nil, fmt.Errorf("no topics configured for lambda %s", lambdaName)
 	}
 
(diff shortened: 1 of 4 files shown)


go/worker/event/kafkaServer.go
```diff
@@ -89,13 +89,15 @@ func (km *KafkaManager) newLambdaKafkaConsumer(consumerName string, lambdaName s
 		return nil, fmt.Errorf("no topics configured for lambda %s", lambdaName)
 	}
 
+	kafkaCfg := common.Conf.Kafka
```
tylerharter, line 92: Aren't we copying the whole struct?  Maybe just point to it instead?

tylerharter (approved):

PR #460, opened 2026-05-21
Containerd sandbox

LeonSun-CS:


go/common/lambdaConfig.go
```diff
@@ -114,6 +114,7 @@ func LoadLambdaConfig(codeDir string) (*LambdaConfig, error) {
 	file, err := os.Open(path)
 
 	if errors.Is(err, os.ErrNotExist) {
+		// TODO: use slog
```
tylerharter, line 117: Probably simpler to just do it rather than add a comment (or just leave it out of this PR since it is not related).

go/worker/lambda/packages/packagePuller.go
```diff
@@ -165,7 +165,8 @@ func (pp *PackagePuller) sandboxInstall(p *Package) (err error) {
 		alreadyInstalled = true
 	} else {
 		slog.Info(fmt.Sprintf("run pip install %s from a new Sandbox to %s on host", p.Name, scratchDir))
-		if err := os.Mkdir(scratchDir, 0700); err != nil {
+		filesDir := filepath.Join(scratchDir, "files")
```
tylerharter, line 168: Why do we need this change?

go/go.mod
```diff
@@ -7,12 +7,15 @@ toolchain go1.24.4
 require (
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.8.2
 	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.6.0
+	github.com/containerd/containerd v1.6.26
 	github.com/fsouza/go-dockerclient v1.10.1
 	github.com/golang-jwt/jwt v3.2.2+incompatible
+	github.com/opencontainers/runtime-spec v1.0.3-0.20210326190908-1c3f411f0417
```
tylerharter, line 13: I think containerd is obvious, but could you let me know what each of the other two dependencies provide?

go/worker/helpers.go
```diff
@@ -373,7 +373,9 @@ func stoppedDirtyToStoppedClean(olPath string) error {
 		// Log an error if unmounting the main directory fails.
 		if errors.Is(err, syscall.EINVAL) {
 			fmt.Printf("Sandbox mount root is not mounted. No need to clean up.\n")
-		} else {
+		} else if errors.Is(err, syscall.ENOENT) {
```
tylerharter, line 376: is this related to the other changes?   As, does in surface more regularly now?  Might be simpler as a separate PR is this is an unrelated issue you stumbled upon.

go/worker/sandbox/dockerPool.go
```diff
@@ -152,8 +152,8 @@ func (pool *DockerPool) Create(parent Sandbox, isLeaf bool, codeDir, scratchDir
 		return nil, err
 	}
 
-	if err := waitForServerPipeReady(c.HostDir()); err != nil {
-		c.Destroy("waitForServerPipeReady failed")
+	if err := WaitForServerPipeReady(c.HostDir()); err != nil {
```
tylerharter, line 155: Is using the pipe the most elegant way to do this?  We don't necessarily need to follow what the docker implementation is doing.

go/worker/sandbox/containerdPool.go
```diff
@@ -0,0 +1,290 @@
...
+	var execProcess containerd.Process
+
+	defer func() {
+		if err != nil {
+			containerdutil.CleanupContainerdResources(ctx, "", container, task, execProcess)
+		}
+	}()
+
+	// Resource limit calculations
+	cpuPeriod := uint64(100000)                                                // using Standard CFS period: 100ms (100,000 microseconds)
```
tylerharter, line 156: Can you just put the comments before?  In a 2-column GitHub diff, it is hard to read this long.

tylerharter (changes requested):
Haven't read everything yet, still working through it.

go/worker/helpers.go
LeonSun-CS: Yes, this is related. The context is when stopped dirty, `stoppedDirtyToStoppedClean` func runs and tries to unmount "root-sandboxes", and containerd would ran into no such file error because it doesn't use "root-sandboxes" dir.

